PR #3, opened 2022-03-02
correct logic on detecting a pinhole reset

geissonator:
Per our circuit designer, a 0 actually is what indicates a pinhole
reset. My testing was in simulation so I was able to falsely test
my thinking of how the hardware behaved.

There is also still an ongoing discussion on why some machines show the
inverse of this logic but the circuit designer assures me that a 1 is
the default and indicates the reboot was not a pinhole reset and a 0
indicates it was a pinhole reset.

Change-Id: I24e536e2dcb8b3263f4fa407bc6f91fdef3ae0ad

### DIFF
--- a/bmc_state_manager.cpp
+++ b/bmc_state_manager.cpp
@@ -306,7 +306,8 @@ void BMC::discoverLastRebootCause()
     auto gpioval =
         phosphor::state::manager::utils::getGpioValue("reset-cause-pinhole");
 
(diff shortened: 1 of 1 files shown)


geissonator:
Upstream commit: https://gerrit.openbmc-project.xyz/c/openbmc/phosphor-state-manager/+/51682

geissonator:
Upstream has the +1's.

PR #4, opened 2022-03-03
Stop FSI scan service if chassis power on but host not booting

eddiejames:
If the host is then started, CFAM reset will start so FSI needs to be
scanned again, otherwise FSI access is broken.


geissonator:
add to approvelist

geissonator:
I opened https://github.com/openbmc/phosphor-state-manager/issues/19 to track an upstream fix for this.

PR #5, opened 2022-03-03
host-sensors: Reset when system off but last state was on

msbarth:
When power was on before a BMC reboot and system is now off, the host
sensors need to be reset to show the correct system state of `off` now.
This is done by starting the `phosphor-reset-sensor-states@0.service`
unit, which handles setting the appropriate dbus properties.

Tested:
  Power on system and then remove/add AC
  Check BootProgress, etc.. properties set by reset service

Change-Id: I4928a231495a1295425dc828606b40de03253bab
Signed-off-by: Matthew Barth <msbarth@us.ibm.com>


msbarth:
upstream review -> https://gerrit.openbmc-project.xyz/c/openbmc/phosphor-state-manager/+/51656

PR #11, opened 2022-04-11
Turn off power-on LEDs after a power blackout

spinler:
After a power blackout, the power-on LED group, which is asserted when
power is on, would have been restored to still be asserted.  Turn it off
explicitly since the code knows power is now off.

This fixes SW547826

Signed-off-by: Matt Spinler <spinler@us.ibm.com>


spinler:
I'm thinking this can't go upstream since that LED group name is configured in a system specific JSON file.

geissonator (approved):
Only concern is another downstream-only patch that some day we'll have to pay the piper on.

PR #12, opened 2022-04-14
host-check: add extra debug and time

geissonator:
Defect SW548561 showed an issue where host-check did not correctly get
the host state from pldm and assumed the system was off. The journal
logs were inconclusive on where the issue was, but at some point later,
it did appear pldmd was up and running and could talk to the host.

Use this commit to add some additional debug and a few more retries in
this path.


geissonator:
Going to wait on sending upstream until we can get some more info on what's happening. By design, the pldmd interface should be available and returning the correct data once pldmd has put itself on d-bus.

spinler (approved):

PR #18, opened 2022-05-18
1020 ghe apr on power restore

geissonator:
Upstream Commits:
- https://gerrit.openbmc-project.xyz/c/openbmc/phosphor-state-manager/+/53687
- https://gerrit.openbmc-project.xyz/c/openbmc/phosphor-state-manager/+/53688

Fixes SW550495


geissonator:
Merged upstream so merging here

PR #19, opened 2022-07-09
delay apr power on by 30 seconds

geissonator:
Recently code was added to move the vpd collection services to later in
the BMC boot. This caused a conflict with the APR (Automatic Power
Restart) service. The APR service would start the power on of the host
before vpd collection had completed. This caused conflicts on the shared
i2c busses between the BMC and host firmware.

Long term, we need a way to know when the vpd collection services have
completed and APR should not run until the BMC has reached READY state
and the vpd collection has completed.

Short term, add a 30 second delay when APR detects it needs to power on
the system. This ensure APR does not impact the BMC boot when it's not
needed and adds a safe delay when it does need to boot the system.

Tested:
- Powered on p10bmc chassis, caused an AC cycle, and verified that APR
  did not initiate the boot until after vpd collection had completed.
  No host i2c errors were logged and system booted as expected.
- Issued a r/r and verified service did not run
- Verified service exited quickly when no host power on was needed

### DIFF
--- a/discover_system_state.cpp
+++ b/discover_system_state.cpp
@@ -18,6 +18,7 @@
 #include <iostream>
 #include <map>
 #include <string>
+#include <thread>
 
 namespace phosphor
 {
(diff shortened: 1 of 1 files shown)


geissonator:
Fixes SW554096
This will not be going upstream, we need a more elegant solution. Probably making this a normal service and waiting for BMC_READY (and also figuring out how to know when vpd collection is "complete")

santoshpuranik (approved):
Just need to add a delay in the "always OFF" case too (since that path also runs a CFAM reset).
Once we test with that added, this looks like it works

PR #20, opened 2022-07-20
host-reboot: enforce host boot count on host crash

geissonator:
The host reboot path can be triggered via paths outside of the normal
RequestedHostTransition property set. For example, if the host crashes,
it just triggers an automatic reboot using the systemd targets, which
means the check to prevent continuous host reboots is not run.

This commit ensures we decrement the reboot counter when a host crash
occurs, and it also ensures we check that count when deciding if a
reboot should occur.

While testing, it was found when we end in the host Quiesce state (after
all reboot attempts have expired), the host stayed in Quiesce, even
after issuing a power off. The fix for that was to ensure the
obmc-host-quiesce@.target conflicted with obmc-host-stop@.target to
ensure it would properly run on the power off request and update the
host state to Off.

Tested:
- Verified that AttemptsLeft was properly decremented on host crashes
- Verified that when AttempsLeft reached 0, the reboot was halted
  and the host state was moved to Quiesced
- Verified that once in Quiesced, system could be powered off, and a
  boot worked fine (and AttempsLeft was back to 3)

Change-Id: I8d101b987517e160becc712ed70950f1d3204397

### DIFF
--- a/host_state_manager.cpp
+++ b/host_state_manager.cpp
@@ -63,6 +63,7 @@ constexpr auto HOST_STATE_DIAGNOSTIC_MODE =
     "obmc-host-diagnostic-mode@0.target";
 
 constexpr auto HOST_STATE_QUIESCE_TGT = "obmc-host-quiesce@0.target";
+constexpr auto HOST_STATE_CRASH_TGT = "obmc-host-crash@0.target";
 
 constexpr auto ACTIVE_STATE = "active";
 constexpr auto ACTIVATING_STATE = "activating";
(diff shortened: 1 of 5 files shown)


NodeMan97 (approved):
Looks good as long as it works. I took some time looking at all the targets/interfaces used and they all appear to be in downstream.
The only thing that really stood out is the 'libexecdir'. It was not present in meson_options.txt, however it seems to be used as a constant across the entirety of the project. 
Also if 'obmc-host-startmin@.target' is hit, it will put a kabosh on the 'obmc-host-reboot@.target', but i'm sure that is intended. 
We're also hitting 'obmc-host-stop@.target' in the reboot service just to stop it once we either go into a quiesced or start-min state. That sounds like what's needed to provide the functionality.
My one question is why do we only care if the host was already in a running state to introduce a quiescence or another reboot? What if the host crashes when it is transitioning to running? Wouldn't that still cause an auto reboot of the system?

geissonator:
> The only thing that really stood out is the 'libexecdir'. It was not present in meson_options.txt, however it seems to be used as a constant across the entirety of the project.

Yep, that's just a meson option to put the script in a directory for executables that you don't want in the normal /usr/bin type paths (i.e. it wont be in your ssh command like path by default)

> Also if 'obmc-host-startmin@.target' is hit, it will put a kabosh on the 'obmc-host-reboot@.target', but i'm sure that is intended.

Yep, the reboot target starts final act is to start the startmin (and end itself)

> My one question is why do we only care if the host was already in a running state to introduce a quiescence or another reboot? What if the host crashes when it is transitioning to running? Wouldn't that still cause an auto reboot of the system?

The MPIPL (host-crash) path is only hit when the host is running. Systemd has some "features" where it will send multiple signals indicating a target has started. This caused us to decrement the reboot counter multiple times throughout the reboot (very bad). So the fix (workaround?) was to verify host state so we for sure only decrement it once per crash.

Thanks for the review!

PR #22, opened 2022-08-19
power-restore: always honor AlwaysOn policy

geissonator:
An internal bug was opened recently noting that when the
only-run-apr-on-power-loss config option was set, the AlwaysOn policy
was not executed. This didn't make sense because the AlwaysOn policy is
specifically saying to always power the system on. This commit changes
the logic to always power on, no matter what, if the user has set the
policy to AlwaysOn.

This change fixes a subtle bug in the one-time logic as well. Prior
to this change, the one-time setting would always be executed even if
the only-run-apr-on-power-loss config option was set. The reasoning
behind the bug was that the one-time did need to always be executed,
when it was set to AlwaysOn. The one-time is used in situations where
host or BMC firmware needs to reboot the BMC but ensure the system boots
back up after the BMC reboot (inband code update, brownout conditions).
This commit ensures we continue to execute that behavior but also
ensures we do not run the one-time otherwise when
only-run-apr-on-power-loss is set and there was not an AC loss.

Tested:
- Verified AlwaysOn always powered on system
- Verified power was only "Restored" when there was an AC loss

Change-Id: Ic5ba84cf2d4fe80fc60ebf9440520a80d91c9ed9

### DIFF
--- a/discover_system_state.cpp
+++ b/discover_system_state.cpp
@@ -108,6 +108,18 @@ int main(int argc, char** argv)
     methodUserSetting.append(powerRestoreIntf, "PowerRestorePolicy");
 
     std::variant<std::string> result;
+    bool acLoss = false;
+    // Only run APR iff chassis power was on prior to
+    // the BMC reboot (unless policy is AlwaysOn)
+    auto size = std::snprintf(nullptr, 0, CHASSIS_LOST_POWER_FILE, 0);
+    size++; // null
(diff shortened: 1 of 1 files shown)


geissonator:
Upstream review: https://gerrit.openbmc.org/c/openbmc/phosphor-state-manager/+/56629

anoo1 (approved):

spinler (approved):

PR #24, opened 2022-10-27
only quiesce in reboot on host crash

geissonator:
This script was introduced to ensure a system properly halted in scenarios where the host has crashed (obmc-host-crash@.target). There was a loophole where the standard reboot attempt logic check was not running in this specific path.

Unfortunately, that original change was not done quite right. This script is actually called on all host reboots so we have run into situations where the host is being put into quiesce before being given one last try to boot. That last try in non-host crash scenarios is where certain recovery actions (like trying alternate boot sources for some devices) is done.

This commit modifies the script to ensure it only moves the host to the quiesce state when the cause of the host reboot was a host crash.

Tested:
- Verified in normal reboot path, even if AttemptsLeft is 0, the reboot is still done
- Verified that if AttemptsLeft is 0 and reboot reason is HostCrash that host was put in quiesce

Change-Id: I18e21cc33b3c38334ad760148a0754e2bd7d0780

### DIFF
--- a/scripts/host-reboot
+++ b/scripts/host-reboot
@@ -16,6 +16,14 @@ get_reboot_count()
         | cut -d ' ' -f2
 }
 
+get_restart_cause()
+{
+    busctl get-property xyz.openbmc_project.State.Host \
+        /xyz/openbmc_project/state/host0 \
+        xyz.openbmc_project.State.Host RestartCause \
+        | cut -d '"' -f2
+}
+
 # host instance id is input parameter to function
 host_quiesce()
 {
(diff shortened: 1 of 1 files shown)


geissonator:
Upstream commit: https://gerrit.openbmc.org/c/openbmc/phosphor-state-manager/+/58243
Fixes: SW557839

geissonator:
Further 1020 testing:
- Issued host reboot 3 times during hostboot portion of boot, verified system booted on alternate SBE side when AttemptsLeft was 0
- Started obmc-host-crash@0.target 3 times during PHYP boot and verified on 3rd time we halted in Quiesce

PR #26, opened 2022-11-07
obmc-host-graceful-quiesce: ensure starts synch targets

geissonator:
As seen in other scenarios, the power off synchronization targets do not always start in time when the obmc-host-graceful-quiesce@.target is started. This can result the host being stopped before it has had the chance to gracefully shut itself down.

Ensure the synchronization targets are started by adding them into the target with 'Wants' relationships.

Tested:
- Booted system into hostboot, manually started the target and verified via the journal that the pldm-softpoweroff application completed prior to instructions stopping.

Change-Id: Iede5194c24ad05467235419eaf0f343bccdffc1a

### DIFF
--- a/target_files/obmc-host-graceful-quiesce@.target
+++ b/target_files/obmc-host-graceful-quiesce@.target
@@ -1,4 +1,7 @@
 [Unit]
 Description=Graceful Quiesce Target
 After=multi-user.target
+Wants=obmc-host-stop-pre@%i.target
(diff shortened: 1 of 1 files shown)


geissonator:
Fixes: PE00DYTH
Upstream Fix: https://gerrit.openbmc.org/c/openbmc/phosphor-state-manager/+/58515

Needs to go in 1020.30 and FW1040 as well.

PR #34, opened 2023-02-16
increase timeout on systemd Subscribe method call

geissonator:
Unfortunately the previous attempt at fixing this via retries was unsuccessful. The retries returned an error saying a reply was pending.

This resulted in me going back to the OP940 workaround that was put in place for this and simply increasing the timeout. The retry seemed more efficient but better to have something that works.

This fix is based on 5cfeb07 without making it configurable as there is no need for that type of feature here.

Tested
- Basic good path testing to ensure no regressions

### DIFF
--- a/utils.cpp
+++ b/utils.cpp
@@ -28,26 +28,21 @@ void subscribeToSystemdSignals(sdbusplus::bus::bus& bus)
 {
     auto method = bus.new_method_call(SYSTEMD_SERVICE, SYSTEMD_OBJ_PATH,
                                       SYSTEMD_INTERFACE, "Subscribe");
-    // There are times during the BMC boot where systemd is unable to respond to
-    // the "Subscribe" call. Handle this by retrying the call up to 3 times
-    // before logging an error
-    for (int i = 0; i < 3; i++)
+
+    try
     {
-        try
-        {
-            bus.call_noreply(method);
-        }
-        catch (const sdbusplus::exception::exception& e)
-        {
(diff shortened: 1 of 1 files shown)


geissonator:
Hot fix so no upstream for now but I'll at least send the simplified util upstream at some point.
Fixes 409858

utils.cpp
```diff
@@ -28,26 +28,21 @@ void subscribeToSystemdSignals(sdbusplus::bus::bus& bus)
...
-                  e);
-            continue;
-        }
-        return;
+        // On OpenBMC based systems, systemd has had a few situations where it
+        // has been unable to respond to this call within the default d-bus
+        // timeout of 25 seconds. This is due to the large amount of work being
+        // done by systemd during OpenBMC startup. Set the timeout for this call
+        // to 60 seconds.
+        bus.call(method, (60 * 1000000L));
```
smccarney, line 39: For the upstream commit, might be clearer to use std::chrono_literals and SdBusDuration instead.

smccarney (approved):

utils.cpp
smccarney: https://github.com/openbmc/phosphor-power/blob/master/phosphor-regulators/src/regsctl/utility.hpp#L31

geissonator: Thanks, will do when I send upstream!

geissonator:
Upstream: https://gerrit.openbmc.org/c/openbmc/phosphor-state-manager/+/61017

PR #57, opened 2024-09-13
RBMC: Choose role based on position

spinler:
At this point, there is only a minimal simulation model, and information about the sibling BMC is not yet available.  What is available is the BMC position, which is set by the simulation startup script by writing a 'bmc_position' udev environment variable:

```
$ fw_printenv bmc_position
bmc_position=0
```

This commit makes use of that variable to determine if it should be Active or Passive, where position zero is for the Active BMC. Eventually this would be replaced by a D-Bus property hosted by some other application.

A Services class was created to obtain that and future system information in a way that can be mocked if necessary for unit testing.

A new role_determination::run() function was created that for now just takes the BMC position value, but in the future will probably need to take many more parameters.  A structure is used for that so that it can be initialized with designated initializers (i.e. .name = value ) to make it obvious which value is which.

Tested:

BMC 0:
```
Role Determined: xyz.openbmc_project.State.BMC.Redundancy.Role.Active
```

BMC 1:
```
Role Determined: xyz.openbmc_project.State.BMC.Redundancy.Role.Passive
```

### DIFF
--- a/redundant-bmc/src/manager.cpp
+++ b/redundant-bmc/src/manager.cpp
@@ -6,9 +6,11 @@
 namespace rbmc
 {
 
-Manager::Manager(sdbusplus::async::context& ctx) :
+Manager::Manager(sdbusplus::async::context& ctx,
+                 std::unique_ptr<Services>&& services) :
     ctx(ctx),
-    redundancyInterface(ctx.get_bus(), RedundancyInterface::instance_path)
+    redundancyInterface(ctx.get_bus(), RedundancyInterface::instance_path),
+    services(std::move(services))
 {
     ctx.spawn(startup());
 }
@@ -17,11 +19,9 @@ Manager::Manager(sdbusplus::async::context& ctx) :
 // NOLINTNEXTLINE
 sdbusplus::async::task<> Manager::startup()
 {
-    // TODO: Actually figure out the role
-    lg2::info("Setting role to passive");
-    redundancyInterface.role(RedundancyInterface::Role::Passive);
-
     ctx.spawn(doHeartBeat());
+
+    redundancyInterface.role(determineRole());
 
     co_return;
 }
@@ -40,6 +40,30 @@ sdbusplus::async::task<> Manager::doHeartBeat()
     }
 
     co_return;
+}
+
+Role Manager::determineRole()
+{
+    Role role{Role::Unknown};
+
+    try
+    {
+        role_determination::Input input{
+            .bmcPosition = services->getBMCPosition()};
+
+        role = role_determination::run(input);
+    }
+    catch (const std::exception& e)
+    {
+        lg2::error("Exception while determining, role.  Will have to be "
+                   "passive. Error = {ERROR}",
+                   "ERROR", e);
+        role = Role::Passive;
+    }
+
+    lg2::info("Role Determined: {ROLE}", "ROLE", role);
+
+    return role;
 }
 
 } // namespace rbmc
--- a/redundant-bmc/test/role_determination_test.cpp
+++ b/redundant-bmc/test/role_determination_test.cpp
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+#include "role_determination.hpp"
+
+#include <gtest/gtest.h>
+
+using namespace rbmc;
+using namespace role_determination;
+
+TEST(RoleDeterminationTest, RoleDeterminationTest)
+{
+    {
+        Input input{.bmcPosition = 0};
+        EXPECT_EQ(run(input), Role::Active);
+    }
+
+    {
+        Input input{.bmcPosition = 1};
+        EXPECT_EQ(run(input), Role::Passive);
+    }
+}
--- a/meson.build
+++ b/meson.build
@@ -286,8 +286,26 @@ build_tests = get_option('tests')
 # If test coverage of source files within the root directory are wanted,
 # need to define and build the tests from here
 if not build_tests.disabled()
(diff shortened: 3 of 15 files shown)


redundant-bmc/test/role_determination_test.cpp
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: Apache-2.0
+#include "role_determination.hpp"
+
+#include <gtest/gtest.h>
+
+using Role =
```
RameshIyyar, line 6: Noticed that this defined many places, would you prefer to move into `role_determination.hpp` and remove in other places?

redundant-bmc/test/role_determination_test.cpp
```diff
@@ -0,0 +1,22 @@
...
+    sdbusplus::common::xyz::openbmc_project::state::bmc::Redundancy::Role;
+
+namespace rd = role_determination;
+
+TEST(RoleDeterminationTest, RoleDeterminationTest)
```
RameshIyyar, line 11: It’s now acceptable to have multiple test cases within a single test, but I believe several factors will determine this approach. We can consider adding multiple test cases by splitting them based on potential future changes.

I am not sure why are you preferring to combine multiple scenarios in the single test case (noticed the same in phosphor-logging as well) , is there any motive behind this?

redundant-bmc/src/manager.cpp
```diff
@@ -42,4 +48,15 @@ sdbusplus::async::task<> Manager::doHeartBeat()
     co_return;
 }
 
+Manager::Role Manager::determineRole()
+{
+    role_determination::Input input{.bmcPosition = services->getBMCPosition()};
```
RameshIyyar, line 53: There is chance to get exception, do you want to handle that or crash application?

redundant-bmc/src/manager.cpp
```diff
@@ -17,9 +21,11 @@ Manager::Manager(sdbusplus::async::context& ctx) :
 // NOLINTNEXTLINE
 sdbusplus::async::task<> Manager::startup()
 {
-    // TODO: Actually figure out the role
-    lg2::info("Setting role to passive");
-    redundancyInterface.role(RedundancyInterface::Role::Passive);
+    ctx.spawn(doHeartBeat());
+
+    auto role = determineRole();
```
RameshIyyar, line 26: We could remove `role` temporary variable and call `determineRole` in the below line.

spinler: I have changes coming in this area in the future, but yea that would be better for now.

redundant-bmc/test/role_determination_test.cpp
spinler: good idea, I'll look into the best place to define this once.

redundant-bmc/src/manager.cpp
spinler: I'll look into it.

spinler: If I catch an exception, I guess about all I can do is become Passive since I won't have enough information to do anything else.  That's probably better that letting it die so there is no Role on dbus/redfish at all. Will fix.

redundant-bmc/test/role_determination_test.cpp
spinler: I'm having a hard time on where there best place would be, since it's going to be used in other files that otherwise wouldn't need role_determination.hpp. 

Do you think a standalone header would make sense for this?  Like a types.hpp?  Otherwise, maybe I'll leave as-is for now.

RameshIyyar: > Do you think a standalone header would make sense for this? Like a types.hpp?

Usually, I will follow this to define types/constants which could be used more than one place or some generic one.

> I'll leave as-is for now.

Sure,

RameshIyyar (approved):

redundant-bmc/test/role_determination_test.cpp
spinler: Waffled a bit on this, and then went with your first suggestion to put it role_determination.hpp.

RameshIyyar (approved):

redundant-bmc/src/manager.cpp
```diff
@@ -42,4 +42,28 @@ sdbusplus::async::task<> Manager::doHeartBeat()
     co_return;
 }
 
+Role Manager::determineRole()
+{
+    Role role{Role::Unknown};
+
+    try
+    {
+        role_determination::Input input{
```
geissonator, line 51: I'm wondering if all the complexity we're bringing along for AIO is really needed in this app? Wont this app mostly be synchronous anyway (i.e. it can't really do anything until it determines role?). I'm just a simple person who  likes simple code. But I think this is also an opportunity to get a good grasp on what we get with the new C++ AIO features.

spinler: You mean the coroutines?  I think they're a good improvement.  You don't have to worry about D-Bus being hung while you're making a D-Bus call to another daemon, and instead of async timer callbacks you can just use while loops.  Matches are also while loops.

geissonator: They make my head hurt though :)

geissonator (approved):

redundant-bmc/src/manager.cpp
spinler: FWIW, these lines right here aren't async.  It's just a normal function.  There will eventually be a half dozen or so structure members, which is why the args aren't just passed in normally.

PR #59, opened 2024-10-09
phosphor-set-chassis-transition: ensure only run once

geissonator:
If a user were to power their system on, and then off, and then off again, they will see their CurrentPowerState set to TransitioningToOff.

Although it's a bit of an unusual use case, it's fairly common in the lab/dev environment when users want to quickly ensure their system is in a powered off or on state (without first looking at its current state).

Similar to what is in the host transition services, ensure we only run the chassis transition services once per power on or off.

Tested:
- Ensured multiple power off requests will not leave CurrentPowerState in TransitioningToOff state

Change-Id: I4040072ed0b42f37c9488994b0f7a5975a2bb7c9

### DIFF
--- a/service_files/phosphor-set-chassis-transition-to-off@.service
+++ b/service_files/phosphor-set-chassis-transition-to-off@.service
@@ -8,6 +8,7 @@ Conflicts=obmc-chassis-poweron@%i.target
 [Service]
 Restart=no
 Type=oneshot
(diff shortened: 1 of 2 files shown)


geissonator:
Upstream commit: https://gerrit.openbmc.org/c/openbmc/phosphor-state-manager/+/75143

PR #60, opened 2024-10-10
RBMC: Start looking at sibling BMC

spinler:
The first commit starts looking at the xyz.openbmc_project.State.BMC.Redundancy.Sibling D-Bus interface.  It creates a class around it that can provide access to the properties.

The second commit starts using the sibling fields in active vs passive role determination.

### DIFF
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -9,13 +9,22 @@ redundancy related properties.
 
 ## Startup
 
-So far on startup the application will just immediately determine its role.
+On startup, the code will wait for up to six minutes for the sibling BMC's
+heartbeat to start, assuming the BMC is present. After that it will determine
+its role.
 
 ## Role Determination Rules
 
 The current rules for role determination are:
 
-1. If BMC position zero choose the active role, otherwise passive.
+1. If the sibling BMC doesn't have a heartbeat, choose active. It could be the
+   sibling isn't even present.
+1. If the sibling BMC's position matches this BMC's position, choose passive.
+   This is an error case.
+1. If the sibling isn't provisioned, choose active.
+1. If the sibling is already passive, choose active.
+1. If the sibling is already active, choose passive.
+1. If this BMC's position is zero choose active, otherwise passive.
 
 If there is an internal failure during role determination, like an exception,
 the BMC will also have to become passive.
--- a/redundant-bmc/src/manager.cpp
+++ b/redundant-bmc/src/manager.cpp
@@ -9,11 +9,14 @@
 namespace rbmc
 {
 
+const std::chrono::minutes siblingTimeout{6};
+
 Manager::Manager(sdbusplus::async::context& ctx,
-                 std::unique_ptr<Services>&& services) :
+                 std::unique_ptr<Services>&& services,
+                 std::unique_ptr<Sibling>&& sibling) :
     ctx(ctx),
     redundancyInterface(ctx.get_bus(), RedundancyInterface::instance_path),
-    services(std::move(services))
+    services(std::move(services)), sibling(std::move(sibling))
 {
     ctx.spawn(startup());
 }
@@ -22,7 +25,14 @@ Manager::Manager(sdbusplus::async::context& ctx,
 // NOLINTNEXTLINE
 sdbusplus::async::task<> Manager::startup()
 {
+    co_await sibling->init();
+
     ctx.spawn(doHeartBeat());
+
+    if (sibling->isBMCPresent())
+    {
+        co_await sibling->waitForSiblingUp(siblingTimeout);
+    }
 
     redundancyInterface.role(determineRole());
 
--- a/redundant-bmc/src/manager.cpp
+++ b/redundant-bmc/src/manager.cpp
@@ -80,26 +80,42 @@ sdbusplus::async::task<> Manager::doHeartBeat()
 
 Role Manager::determineRole()
 {
-    Role role{Role::Unknown};
+    using namespace role_determination;
+
+    RoleInfo roleInfo{Role::Unknown, ErrorCase::noError};
 
     try
     {
-        role_determination::Input input{
-            .bmcPosition = services->getBMCPosition()};
+        // Note:  If these returned nullopts, the algorithm wouldn't use
+        //        them anyway because there would be no heartbeat.
+        auto siblingRole = sibling->getRole().value_or(Role::Unknown);
+        auto siblingProvisioned = sibling->getProvisioned().value_or(false);
+        auto siblingPosition = sibling->getPosition().value_or(0xFF);
 
-        role = role_determination::run(input);
+        role_determination::Input input{
+            .bmcPosition = services->getBMCPosition(),
+            .siblingPosition = siblingPosition,
+            .siblingRole = siblingRole,
+            .siblingHeartbeat = sibling->hasHeartbeat(),
+            .siblingProvisioned = siblingProvisioned};
+
+        roleInfo = role_determination::run(input);
     }
     catch (const std::exception& e)
     {
         lg2::error("Exception while determining, role.  Will have to be "
                    "passive. Error = {ERROR}",
                    "ERROR", e);
-        role = Role::Passive;
+        roleInfo.role = Role::Passive;
+        roleInfo.error = ErrorCase::internalError;
     }
 
-    lg2::info("Role Determined: {ROLE}", "ROLE", role);
+    if (roleInfo.error != role_determination::ErrorCase::noError)
+    {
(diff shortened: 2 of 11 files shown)


redundant-bmc/src/sibling_impl.cpp
```diff
@@ -0,0 +1,293 @@
...
+    {
+        lg2::info("Sibling::init called more than once");
+        co_return;
+    }
+
+    ctx.spawn(watchInterfaceAdded());
+    ctx.spawn(watchInterfaceRemoved());
+    ctx.spawn(watchPropertyChanged());
+
+    serviceName = co_await getServiceName();
```
geissonator, line 30: A few comments throughout this function would be appreciated

redundant-bmc/src/sibling_impl.cpp
```diff
@@ -0,0 +1,293 @@
...
+                size_t count = 0;
+                do
+                {
+                    using namespace std::chrono_literals;
+                    co_await sdbusplus::async::sleep_for(ctx, 100ms);
+                    serviceName = co_await getServiceName();
+                    lg2::info(
+                        "After interfacesAdded, sibling service is {SERVICE}",
+                        "SERVICE", serviceName);
+                } while (serviceName.empty() && (count++ < 200));
```
geissonator, line 181: Always been a fan of more descriptive constants for things like this (200).

redundant-bmc/README.md
```diff
@@ -9,13 +9,23 @@ redundancy related properties.
...
 
 1. If BMC position zero choose the active role, otherwise passive.
+1. If the sibling BMC doesn't have a heartbeat, choose active. It could be the
+   sibling isn't even present.
+1. If the sibling BMC's position matches this BMC's position, choose passive.
+   This is an error case.
+1. If the sibling isn't provisioned, choose active.
+1. If the sibling is already passive, choose active.
+1. If the sibling is already active, choose passive.
+1. If this BMC's position is zero choose active, otherwise passive.
```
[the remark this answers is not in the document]
spinler: done

redundant-bmc/src/manager.cpp
```diff
@@ -70,26 +80,42 @@ sdbusplus::async::task<> Manager::doHeartBeat()
...
+
+    RoleInfo roleInfo{Role::Unknown, ErrorCase::noError};
 
     try
     {
+        // Note:  If these returned nullopts, the algorithm wouldn't use
+        //        them anyway because there would be no heartbeat.
+        auto siblingRole = sibling->getRole().value_or(Role::Unknown);
+        auto siblingProvisioned = sibling->getProvisioned().value_or(true);
+        auto siblingPosition = sibling->getPosition().value_or(0xFF);
```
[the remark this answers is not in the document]
spinler: getPosition() returns a size_t (as an optional).  The only reason it would return the nulltopt is if the sibling isn't responding, in which case we couldn't get far enough to compare positions anyway.  What do you think i should do here?

redundant-bmc/src/sibling_impl.cpp
spinler: done

redundant-bmc/src/sibling_impl.cpp
spinler: done.

redundant-bmc/src/manager.cpp
```diff
@@ -22,8 +25,15 @@ Manager::Manager(sdbusplus::async::context& ctx,
 // NOLINTNEXTLINE
 sdbusplus::async::task<> Manager::startup()
 {
+    co_await sibling->init();
+
     ctx.spawn(doHeartBeat());
 
+    if (sibling->isBMCPresent())
```
RameshIyyar, line 32: Don't we need to wait for the sibling BMC to be detected if it's not present? I believe both BMCs initialize in parallel, correct? Are you planning to handle this error separately?

redundant-bmc/src/manager.cpp
```diff
@@ -80,26 +80,42 @@ sdbusplus::async::task<> Manager::doHeartBeat()
...
+    using namespace role_determination;
+
+    RoleInfo roleInfo{Role::Unknown, ErrorCase::noError};
 
     try
     {
+        // Note:  If these returned nullopts, the algorithm wouldn't use
+        //        them anyway because there would be no heartbeat.
+        auto siblingRole = sibling->getRole().value_or(Role::Unknown);
+        auto siblingProvisioned = sibling->getProvisioned().value_or(true);
```
RameshIyyar, line 92: Why are we treating the default state as 'Provisioned' for the sibling BMC?

How do we determine if the sibling BMC is not present, not detected, or has timed out, in order to switch to non-redundant mode?

redundant-bmc/src/role_determination.hpp
```diff
@@ -18,16 +18,42 @@ namespace role_determination
...
+    samePositions
+};
+
+/**
+ * @brief The role and the error reason returned from run()
+ */
+struct RoleInfo
+{
+    Role role;
+    ErrorCase error;
```
RameshIyyar, line 43: Do you think adding an error message will be helpful when creating the PEL in the future? Or would capturing the last N journal traces from the RBMC and CFAM manager be more useful for error logging?

(I noticed you have some good traces from the role determination function, and since this is related to error handling, we could consider it in the future to assist with debugging).

redundant-bmc/src/manager.cpp
```diff
@@ -9,11 +9,14 @@
 namespace rbmc
```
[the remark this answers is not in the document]
spinler: LOL I had quite the typo there.

redundant-bmc/src/manager.cpp
spinler: There's going to be a GPIO for hardware presence that says if the BMC card is even plugged in.  So if the card isn't plugged in, there's no point in calling waitForSiblingUp.

If that other sled is plugged in and the sibling does eventually come up before we power on, the plan is to enable redundancy at that point.

redundant-bmc/src/manager.cpp
spinler: It won't matter much because we'll know from the lack of the sibling heartbeat that the sibling isn't alive, but I can change the default to false.

redundant-bmc/src/role_determination.hpp
spinler: I have an upcoming commit to do that, it just didn't get pushed up in this PR.  It'll look like:

```
  if (!result && !input.siblingProvisioned)
    {
        result = {Role::Active, ErrorCase::noError,
                  "Sibling is not provisioned"};
    }

```

and then it will be saved in the journal and also in a persistent data file for debug.  I also have an upcoming 'rbmctool' which will print it.

redundant-bmc/src/sibling_impl.cpp
```diff
@@ -0,0 +1,302 @@
...
+            lg2::info("Waiting for sibling interface and/or heartbeat: "
+                      "Present = {PRES}, Heartbeat = {HB}",
+                      "PRES", interfacePresent, "HB", heartbeat);
+            waiting = true;
+        }
+
+        co_await sdbusplus::async::sleep_for(ctx, 500ms);
+    }
+
+    lg2::info(
```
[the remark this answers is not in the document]
spinler: Yea.  I'm going to handle enabling redundancy kinda like determining the role, where it'll pass in all of  the relevant parameters to a function that will spit out if redundancy can be enabled and the reasons why it can't be if not.

redundant-bmc/src/manager.cpp
spinler: done

redundant-bmc/src/manager.cpp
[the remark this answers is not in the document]
geissonator: Was just hoping the 0xFF could be replaced with Role::Unknown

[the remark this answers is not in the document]
geissonator: Ahh, I see, there is no enum for Position. Oh well.

redundant-bmc/src/role_determination.hpp
RameshIyyar: those debug data is not planned to capture in the PEL if we are planning to create for some error case? Just in journal and a separate persistent file which will be collected as part of the BMC dump?

spinler: I think I have notes that we also have to create a PEL if we can't enable redundancy, so I may as well put the reasons in there too.

RameshIyyar (approved):

geissonator (approved):

PR #63, opened 2024-12-03
RBMC: Add rbmctool executable

spinler:
This tool can display information about and eventually modify redundant BMC behavior.  The idea is it can be used for lab debug, field debug via capturing output in BMC dumps, and allow changing redundancy behavior for bringup and manufacturing test.

The tool uses a new Services::getFWVersion() function that uses the same hash of the VERSION_ID field in /etc/os-release as used in the sibling application.  This is included so that the version of both the local and sibling BMC can be shown together.  In the future this value will also be used when determining if redundancy can be enabled.

So far, the tool provides:

```
$ rbmctool -h
RBMC Tool
Usage: ./rbmctool [OPTIONS]

Options:
  -h,--help                   Print this help message and exit
  -d                          Display RBMC Information
  -e Needs: -d                Add extended RBMC details to the display
```

The -d option shows basic information about the local and sibling BMCs:

```
$ rbmctool -d

Local BMC
-----------------------------
Role:                Active
BMC Position:        0
Redundancy Enabled:  true

Sibling BMC
-----------------------------
Role:                Passive
```

The -e displays extended details:

```
$ rbmctool -de

Local BMC
-----------------------------
Role:                Active
BMC Position:        0
Redundancy Enabled:  true
Failovers Paused:    false
FW version hash:     47048280
Provisioned:         true

Sibling BMC
-----------------------------
Role:                Passive
BMC Position:        1
BMC State:           Ready
Redundancy Enabled:  true
Failovers Paused:    false
Sibling Comm OK:     true
FW version hash:     47048280
Provisioned:         true
```

Eventually more functionality can be added to the tool, such as:

- Force redundancy to be disabled
- Trigger a failover
- Display the details on the role and redundancy-enabled choices

### DIFF
--- a/redundant-bmc/src/rbmc_tool.cpp
+++ b/redundant-bmc/src/rbmc_tool.cpp
@@ -1,0 +1,213 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#include "services_impl.hpp"
+
+#include <CLI/CLI.hpp>
+#include <xyz/openbmc_project/ObjectMapper/client.hpp>
+#include <xyz/openbmc_project/State/BMC/Redundancy/Sibling/client.hpp>
+#include <xyz/openbmc_project/State/BMC/Redundancy/client.hpp>
+#include <xyz/openbmc_project/State/BMC/client.hpp>
+
+#include <format>
+#include <iostream>
+
+using Sibling =
+    sdbusplus::client::xyz::openbmc_project::state::bmc::redundancy::Sibling<>;
+using Redundancy =
+    sdbusplus::client::xyz::openbmc_project::state::bmc::Redundancy<>;
+using BMCState = sdbusplus::client::xyz::openbmc_project::state::BMC<>;
+using Role = Redundancy::Role;
+
+// NOLINTNEXTLINE
+sdbusplus::async::task<std::string> getBMCState(sdbusplus::async::context& ctx)
+{
+    using ObjectMapper =
+        sdbusplus::client::xyz::openbmc_project::ObjectMapper<>;
+    using StateMgr = sdbusplus::client::xyz::openbmc_project::state::BMC<>;
+
+    try
+    {
+        auto mapper = ObjectMapper(ctx)
+                          .service(ObjectMapper::default_service)
+                          .path(ObjectMapper::instance_path);
+
+        std::string statePath = std::string{BMCState::namespace_path::value} +
+                                '/' + BMCState::namespace_path::bmc;
+        std::vector<std::string> stateIface{BMCState::interface};
+
+        auto object = co_await mapper.get_object(statePath, stateIface);
+        auto service = object.begin()->first;
+
+        auto stateMgr = StateMgr(ctx).service(service).path(statePath);
+        auto bmcState = co_await stateMgr.current_bmc_state();
+
+        auto stateString = BMCState::convertBMCStateToString(bmcState);
+        co_return stateString.substr(stateString.find_last_of('.') + 1);
+    }
+    catch (const sdbusplus::exception_t& e)
+    {
+        // Just show the error in the state field
+        co_return e.what();
+    }
+}
+
+// NOLINTBEGIN
+sdbusplus::async::task<>
+    displayLocalBMCInfo(sdbusplus::async::context& ctx, bool extended)
+// NOLINTEND
+{
+    auto redundancy = sdbusplus::async::proxy()
+                          .service("xyz.openbmc_project.State.BMC.Redundancy")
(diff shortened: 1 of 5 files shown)


redundant-bmc/src/rbmc_tool.cpp
```diff
@@ -0,0 +1,174 @@
...
+        std::cout << std::format("Role:                {}\n", role);
+
+        rbmc::ServicesImpl services{ctx};
+        std::cout << std::format("BMC Position:        {}\n",
+                                 services.getBMCPosition());
+
+        auto enabled = std::get<bool>(props.at("RedundancyEnabled"));
+        std::cout << std::format("Redundancy Enabled:  {}\n", enabled);
+
+        if (extended)
```
RameshIyyar, line 51: Do you think adding all CFAM fields for the local BMC, similar to how it was done for the sibling BMC, would be helpful for the BMC dump?

spinler: We can get the actual CFAM fields for the dump by using cfamstool, which has them for the local BMC (and sibling).

I guess here the local BMC doesn't have the BMC state and SiblingComm fields.  I didn't add the BMC state since the code had no use for it yet, but yea I might as well add it and the other one so it has them all.

spinler: Added BMC state to the output.  Let me know if you were meaning something else.

RameshIyyar: I thought dumping all these [fields](https://github.com/ibm-openbmc/openpower-proc-control/tree/1110/rbmc-cfam-daemon#cfam-s-scratchpad-register-layout) for both BMCs would be helpful for debugging. However, we can always add it later if needed.

RameshIyyar (approved):

redundant-bmc/src/rbmc_tool.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+}
+
+// NOLINTBEGIN
+sdbusplus::async::task<>
+    displayLocalBMCInfo(sdbusplus::async::context& ctx, bool extended)
+// NOLINTEND
+{
+    auto redundancy = sdbusplus::async::proxy()
+                          .service(Redundancy::interface)
```
geissonator, line 60: Would hard coding the service name be better then assuming it's the interface name?

redundant-bmc/src/rbmc_tool.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+    try
+    {
+        auto props =
+            co_await redundancy
+                .get_all_properties<Redundancy::PropertiesVariant>(ctx);
+
+        auto role =
+            Redundancy::convertRoleToString(std::get<Role>(props.at("Role")));
+        role = role.substr(role.find_last_of('.') + 1);
```
geissonator, line 75: Could you add a comment on what we're getting back to help explain the "find_last_of(.)"?

redundant-bmc/src/services_impl.cpp
```diff
@@ -155,4 +157,37 @@ bool ServicesImpl::getProvisioned() const
...
+    }
+
+    if (version.empty())
+    {
+        lg2::error("Unable to parse VERSION_ID out of /etc/os-release");
+        // let it hash the empty string
+    }
+
+    return std::format(
+        "{:X}", static_cast<uint32_t>(std::hash<std::string>{}(version)));
```
geissonator, line 190: Hmmm. "std::hash does not guarantee that the result will be the same across computers, builds, or even executions of the same build on the same computer. Your only guarantee is that during one execution, objects that are equal have the same hash."
Not sure we can rely on std::hash to provide a consistent hash across both BMC's.

spinler: It has to be the same in both this app and the CFAM daemon.  So far it's been working fine but I do have a  story open for using SHA512 in both like software-updater uses for its hashes.

So if possible I'd like to leave this as it is for now and then change both the SHA512 before they're actually used for comparing as part of enabling redundancy.

spinler: nevermind, will fix now.

redundant-bmc/src/rbmc_tool.cpp
spinler: done

redundant-bmc/src/rbmc_tool.cpp
spinler: done

redundant-bmc/src/services_impl.cpp

geissonator (approved):

PR #64, opened 2024-12-11
RBMC: Determine the reasons for role and lack of redundancy

spinler:
The main features are:
1. If both BMCs get to role determination at exactly the same time, let the one that was previously passive go first.
2. Save a text reason for why the role was determined, for use in debug.
3. Create new function to determine if redundancy can be enabled, and if not give the reasons why.  It isn't being called yet.


redundant-bmc/README.md
```diff
@@ -36,6 +36,9 @@ The current rules for role determination are:
    just due to an error.
 1. Finally, if this BMC's position is zero choose active, otherwise passive.
 
+The actual reason the code used to determine the role is saved in the
```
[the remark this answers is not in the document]
spinler: It will be available in rbmctool so that it will be in dumps for IBM debug, but otherwise it hasn't been a requirement to make this available on redfish.  We can revisit in the future if necessary.

redundant-bmc/src/role_determination.cpp
```diff
@@ -2,67 +2,68 @@
...
 
-#include <phosphor-logging/lg2.hpp>
+#include <format>
 
 namespace rbmc::role_determination
 {
 
 RoleInfo run(const Input& input)
 {
+    std::optional<RoleInfo> result;
```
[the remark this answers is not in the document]
spinler: Good catch, was left over from a previous version where I was using optionals.  Will fix.

RameshIyyar (approved):

redundant-bmc/test/redundancy_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+    {
+        auto input = golden;
+        input.codeVersionsMatch = false;               // codeMismatch
+        input.siblingState = rbmc::BMCState::Quiesced; // siblingHealth
+        input.siblingHasSiblingComm = false;           // systemHealth
+        input.siblingRole = rbmc::Role::Unknown;       // siblingHealth
+
+        auto reasons = getNoRedundancyReasons(input);
+
+        EXPECT_TRUE(std::ranges::contains(reasons, codeMismatch));
```
geissonator, line 130: Verify size is 4?

redundant-bmc/src/manager.cpp
```diff
@@ -187,8 +186,8 @@ sdbusplus::async::task<std::optional<role_determination::RoleInfo>>
     // An unprovisioned BMC cannot be active.
     if (!services->getProvisioned())
     {
-        lg2::info("Role = passive because BMC not provisioned");
-        co_return RoleInfo{Role::Passive, ErrorCase::notProvisioned};
+        co_return RoleInfo{Role::Passive, ErrorCase::notProvisioned,
+                           "BMC not provisioned"};
```
geissonator, line 190: Can we use getNoRedundancyDescription() to keep the strings all in one place?

redundant-bmc/test/redundancy_test.cpp
spinler: Will do.

redundant-bmc/src/manager.cpp
spinler: Yea, I can change it to do it the same way as the RedundancyReasons was done.

redundant-bmc/src/redundancy.hpp
```diff
@@ -0,0 +1,61 @@
...
+    siblingNoCommunication,
+    codeMismatch,
+    siblingNotAtReady,
+    systemHardwareConfigIssue,
+    other
+};
+
+using NoRedundancyReasons = std::set<NoRedundancyReason>;
+
+NoRedundancyReasons getNoRedundancyReasons(const Input& input);
```
spinler, line 56: forgot the prologue on these.

spinler: fixed.

redundant-bmc/src/manager.cpp
spinler: Simplified things to just have the Role and the RoleReason enum, here, and then added function to get the description and if the reason is an error case.

geissonator (approved):

PR #67, opened 2025-01-22
1110 clang19

geissonator:


redundant-bmc/src/persistent_data.hpp
```diff
@@ -53,6 +53,7 @@ void writeFile(const nlohmann::json& json, const std::filesystem::path& path);
  * @param[in] name - The key to save the value under
  * @param[in] value - The value to save
  */
+// NOLINTBEGIN
```
spinler, line 56: what happened here?

spinler (approved):

redundant-bmc/src/persistent_data.hpp
geissonator: It wanted this, which made no sense, and didn't compile:
```
+++ b/redundant-bmc/src/persistent_data.hpp
@@ -54,7 +54,7 @@ void writeFile(const nlohmann::json& json, const std::filesystem::path& path);
  * @param[in] value - The value to save
  */
 template <typename T>
-void write(std::string_view name, T value,
+void write(std::string_view name, const c& onst c& onst T& value,
            const std::filesystem::path& path = dataFile)
 {
     auto json = util::readFile(path).value_or(nlohmann::json::object());
@@ -64,7 +64,7 @@ void write(std::string_view name, T value,
     }
     else
     {
-        json[name] = value;
+        json[name] = false;
     }

     util::writeFile(json, path);
@@ -118,5 +118,6 @@ std::optional<T> read(std::string_view name,
  */
 void remove(std::string_view name,
             const std::filesystem::path& path = dataFile);
+;
```

PR #74, opened 2025-02-24
RBMC: Take actions on passive BMC failures.

spinler:
The first commit when will disable redundancy when the passive BMC goes to the Quiesce state.  We could still keep syncs running here if we wanted to.

The second commit disables redundancy when contact with the passive is lost for 5 minutes.  Anytime the passive come back, even if less than 5 minutes, there will be an attempt to re-enabled redundancy, which would include a full sync when available.

### DIFF
--- a/redundant-bmc/src/active_role_handler.cpp
+++ b/redundant-bmc/src/active_role_handler.cpp
@@ -36,7 +36,18 @@ sdbusplus::async::task<> ActiveRoleHandler::start()
 
     // TODO: Create an error if no redundancy
 
+    startSiblingWatches();
+
     co_return;
+}
+
+void ActiveRoleHandler::siblingStateChange(BMCState state)
+{
+    if (state == BMCState::Quiesced)
+    {
+        lg2::error("Sibling BMC went to Quiesce, disabling redundancy");
+        redMgr.determineAndSetRedundancy();
+    }
 }
 
 } // namespace rbmc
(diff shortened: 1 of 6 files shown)


redundant-bmc/src/active_role_handler.cpp
```diff
@@ -36,6 +37,67 @@ sdbusplus::async::task<> ActiveRoleHandler::start()
...
+{
+    lg2::info("Passive BMC heartbeat started");
+
+    stopSiblingWatches();
+
+    co_await sdbusplus::async::execution::when_all(
+        sibling.waitForSiblingRole(), sibling.waitForBMCSteadyState());
+
+    // Try to re-enable redundancy if it isn't
+    if (!redundancyInterface.redundancy_enabled())
```
spinler, line 91: I think I need to call it even if it was enabled, because it won't necessarily be able to be enabled again.

redundant-bmc/src/sibling.hpp
```diff
@@ -145,13 +146,24 @@ class Sibling
     void clearCallbacks(Role role)
     {
         redEnabledCBs.erase(role);
+        bmcStateCBs.erase(role);
```
geissonator, line 150: Never a fan of the same thing in two places (in case one changes and we forget the other). Can we just have this call the new clearBMCStateCallback() function?

spinler: Fixed.  Also added a clearHeartbeatCallback() that works the same way.

geissonator (approved):

PR #80, opened 2025-03-25
RBMC: Move getBMCState call into Services

spinler:
Move this function from rbmc_tool.cpp into the Services class so that the application code can call it.  It will be used in the future when checking to see if a failover is possible.

Tested:
rbmctool still works

### DIFF
--- a/redundant-bmc/src/services_impl.cpp
+++ b/redundant-bmc/src/services_impl.cpp
@@ -5,6 +5,7 @@
 
 #include <phosphor-logging/lg2.hpp>
 #include <xyz/openbmc_project/ObjectMapper/client.hpp>
+#include <xyz/openbmc_project/State/BMC/client.hpp>
 #include <xyz/openbmc_project/State/Host/client.hpp>
 
 #include <fstream>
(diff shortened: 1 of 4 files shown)


geissonator (approved):

redundant-bmc/src/services_impl.cpp
```diff
@@ -331,4 +332,27 @@ bool ServicesImpl::isPoweredOn() const
...
+
+    auto mapper = ObjectMapper(ctx)
+                      .service(ObjectMapper::default_service)
+                      .path(ObjectMapper::instance_path);
+
+    std::string statePath = std::string{StateMgr::namespace_path::value} + '/' +
+                            StateMgr::namespace_path::bmc;
+    std::vector<std::string> stateIface{StateMgr::interface};
+
+    auto object = co_await mapper.get_object(statePath, stateIface);
```
RameshIyyar, line 351: Suggestion: Would it be beneficial to add a separate helper API for calling the GetObject DBus API? I noticed that the same API is already being called in the readHostState function but, its your wish.

RameshIyyar (approved):

redundant-bmc/src/services_impl.cpp

RameshIyyar (approved):

PR #81, opened 2025-03-27
RBMC: Document data sync interaction

spinler:
Explain how the code interacts with the data sync daemon.

Change-Id: I28a5075275c98940e454c6d99b46de509f26e5b7

### DIFF
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -121,6 +121,54 @@ Note that redundancy cannot be enabled at runtime if the system wasn't booted
 with redundancy enabled. A concurrent maintenance operation would be necessary
 in that case.
 
+## Interacting with Data Sync on the Active BMC
+
+### When Enabling Redundancy
+
+Any time redundancy is being enabled on the active BMC, or when the active BMC
+detects that the passive BMC recovered from some interruption even if redundancy
+wasn't disabled on D-Bus yet, the following steps will be taken:
+
+1. Set the `DisableSync` property to false.
+1. Call `StartFullSync` to start the full sync.
+1. Wait for the `FullSyncStatus` property change to `FullSyncCompleted` or
+   `FullSyncFailed`.
+
+If the full sync fails, redundancy will be disabled and `DisableSync` will be
+set to true. Note that if the system ever goes through a transition that causes
+another attempt at enabling redundancy, it won't be prevented and the full sync
+will be attempted again.
+
+### When disabling redundancy
+
+Whenever the active BMC attempts to enable redundancy and it can't for some
+reason, it will set the `DisableSync` property to true to stop background
+syncing if it was occurring.
+
+### Background Sync errors
+
+The `SyncEventsHealth` property says if any background sync operations had
+failures. Since retries are built into the sync daemon, it should take something
+serious for the health property to change to critical, such as:
+
+- rsync daemon on either BMC crashing
+- passive BMC reboot
+- loss of network connectivity
+
+Depending on when a sync happens to occur in relation to the root cause of the
+fail, the property change may be the first indication that something is wrong,
+or it may not change until quite some time later if at all.
+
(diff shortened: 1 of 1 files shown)


redundant-bmc/README.md
```diff
@@ -118,3 +118,51 @@ wasn't running.
...
+
+Depending on when a sync happens to occur in relation to the root cause of the
+fail, the property change may be the first indication that something is wrong,
+or it may not change until quite some time later if at all.
+
+To deal with all of this, code will watch for the health property to change to
+critical, and then delay for 5 seconds to allow a sibling heartbeat loss to get
+noticed by the BMC.
+
+After the 5 seconds, if the sibling doesn't have a heartbeat then the code will
```
RameshIyyar, line 164: Why are you preferring this to have two different paths?

If the sync fails, let's disable redundancy.

If we lose the heartbeat, it will trigger the existing handler, which might detect that redundancy is already disabled. However, it could also initiate additional service actions, such as handling a sibling loss due to reasons like a broken CFAM connection or other issues unrelated to sync failures."

It looks like we are waiting for 5 seconds to disable redundancy if the sync fails. I'm trying to understand the reasoning behind this delay.

spinler: From the 'Passive Heartbeat Changes' section above:

> When the passive heartbeat stops, redundancy is functionally disabled as there
is no passive BMC alive to handle the failover method. The active BMC will not
do anything for five minutes to allow time for the passive BMC to come back
after a reboot. At five minutes, it is assumed that the BMC won't come back and
RedundancyEnabled will be set to false and an event log will be created.

We've discussed a few times in the meetings that Dean doesn't want redundancy to go to disabled if the passive BMC reboots.

redundant-bmc/README.md
```diff
@@ -118,3 +118,51 @@ wasn't running.
 Note that redundancy cannot be enabled at runtime if the system wasn't booted
 with redundancy enabled. A concurrent maintenance operation would be necessary
 in that case.
+
+## Interacting with Data Sync on the Active BMC
+
+### When Enabling Redundancy
+
+Any time redundancy is being enabled on the active BMC, or when the active BMC
+detects that the passive BMC had some interruption even if redundancy wasn't
```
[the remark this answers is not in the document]
spinler: good idea.  done.

redundant-bmc/README.md
```diff
@@ -118,3 +118,51 @@ wasn't running.
...
+   `FullSyncFailed`.
+
+If the full sync fails, redundancy will be disabled and `DisableSync` will be
+set to true. Note that if the system ever goes through a transition that causes
+another attempt at enabling redundancy, it won't be prevented and the full sync
+will be attempted again.
+
+### When disabling redundancy
+
+Whenever the active BMC attempts to enable redundancy and it can't for some
```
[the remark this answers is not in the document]
RameshIyyar: As discussed, we're currently only syncing non-critical data—specifically debug data (error logs) from passive to active. Therefore, we don't need to worry about disabling redundancy in case the sync from passive to active fails. We can revisit this if needed in the future.

So I am fine with now.

redundant-bmc/README.md
RameshIyyar: I think I may not have explained my point clearly. Do we really need delay to disable redundancy if the sync fails? Sync failures could occur from Active to Passive as well.

(Sorry I am not sure how I missed this reply notification)

RameshIyyar (approved):

geissonator (approved):

redundant-bmc/README.md
spinler: I will resolve this for now, and when I follow up with code we can revisit if anything needs changing.

spinler:
The last push was just a necessary rebase to fix a merge conflict.  Will now merge.

PR #82, opened 2025-04-01
Don't enable redundancy at runtime if it wasn't enabled when runtime was first reached.

spinler:
Whenever an attempt to enable redundancy at runtime is made, redundancy can't be enabled if it wasn't already enabled when runtime was first reached.  An example of when the check would be done is after either BMC comes back from a reboot at runtime.   Implement this check by:

1. Start tracking if  the system is at off/booting/runtime/other.  While this commit only cares about off vs runtime, there will be uses for the other states in the future, such as pausing failovers during those periods.
2.  When the state reaches runtime, save if redundancy is off and also save a 'valid' flag with it.  When the system goes to Off, clear the valid flag so that the code knows the redundancy value isn't usable until the next time runtime is reached.
3. When redundancy is being determined, also check these 2 new fields as "valid && redundancy off", and don't enable redundancy if both are true.

### DIFF
--- a/redundant-bmc/src/services_impl.cpp
+++ b/redundant-bmc/src/services_impl.cpp
@@ -6,6 +6,7 @@
 #include <phosphor-logging/lg2.hpp>
 #include <xyz/openbmc_project/ObjectMapper/client.hpp>
 #include <xyz/openbmc_project/State/BMC/client.hpp>
+#include <xyz/openbmc_project/State/Boot/Progress/client.hpp>
 #include <xyz/openbmc_project/State/Host/client.hpp>
 
 #include <fstream>
@@ -15,10 +16,13 @@ namespace rbmc
 
 using HostState = sdbusplus::client::xyz::openbmc_project::state::Host<>;
 using ObjectMapper = sdbusplus::client::xyz::openbmc_project::ObjectMapper<>;
+using BootProgress =
+    sdbusplus::client::xyz::openbmc_project::state::boot::Progress<>;
 
 using HostProperties =
     std::variant<std::string, HostState::HostState, HostState::RestartCause,
-                 HostState::Transition, std::set<HostState::Transition>>;
+                 HostState::Transition, std::set<HostState::Transition>,
+                 BootProgress::ProgressStages>;
 using HostPropMap = std::unordered_map<std::string, HostProperties>;
 using HostInterfaceMap = std::map<std::string, HostPropMap>;
 
@@ -47,13 +51,6 @@ constexpr auto systemd = "org.freedesktop.systemd1";
 namespace util
 {
 
-bool getPoweredOnValue(HostState::HostState state)
-{
-    // For the current purposes, consider all other values
-    // as On. May need to revisit in the future.
-    return state != HostState::HostState::Off;
-}
-
 // NOLINTNEXTLINE
 sdbusplus::async::task<std::string> getService(sdbusplus::async::context& ctx,
                                                const std::string& path,
@@ -74,8 +71,12 @@ sdbusplus::async::task<std::string> getService(sdbusplus::async::context& ctx,
 sdbusplus::async::task<> ServicesImpl::init()
 {
     ctx.spawn(watchHostInterfacesAdded());
-    ctx.spawn(watchHostPropertiesChanged());
+    ctx.spawn(watchHostStatePropertiesChanged());
+    ctx.spawn(watchBootProgressPropertiesChanged());
     co_await readHostState();
+    co_await readBootProgress();
+    updateSystemState();
+
     co_return;
 }
 
@@ -86,13 +87,12 @@ sdbusplus::async::task<> ServicesImpl::readHostState()
     {
         auto service = co_await util::getService(ctx, object_path::hostState,
                                                  HostState::interface);
-        auto stateMgr =
-            HostState(ctx).service(service).path(object_path::hostState);
+        hostState = co_await HostState(ctx)
+                        .service(service)
+                        .path(object_path::hostState)
+                        .current_host_state();
 
-        auto hostState = co_await stateMgr.current_host_state();
-        poweredOn = util::getPoweredOnValue(hostState);
-
-        lg2::debug("initial host state is {STATE}", "STATE", hostState);
+        lg2::debug("initial host state is {STATE}", "STATE", hostState.value());
     }
     catch (const sdbusplus::exception_t& e)
     {
@@ -110,6 +110,8 @@ sdbusplus::async::task<> ServicesImpl::watchHostInterfacesAdded()
 
     while (!ctx.stop_requested())
     {
+        bool changed = false;
+
         auto [_, interfaces] =
             co_await match
                 .next<sdbusplus::message::object_path, HostInterfaceMap>();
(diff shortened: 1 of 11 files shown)


redundant-bmc/src/services_impl.cpp
```diff
@@ -142,16 +161,119 @@ sdbusplus::async::task<> ServicesImpl::watchHostPropertiesChanged()
...
+    {
+        newState = SystemState::off;
+    }
+    else if (hostState == HState::TransitioningToRunning)
+    {
+        newState = SystemState::booting;
+    }
+    else if (hostState == HState::Running)
+    {
+        if ((bootProgress.value() == BProgress::SystemInitComplete) ||
```
geissonator, line 248: The only other one which can sort of mean runtime is "SystemSetup". This is where we go when PHYP is booted but it's NVRAM has been cleared and you need to re-initialize it in the HMC. My vote would be we don't worry about it, but figured I'd put it out there in case you see something you want to handle there.

geissonator (approved):

RameshIyyar (commented):
> This is because the host may treat the passive BMC differently if it
isn't enabled, and so the host to passive BMC link may not be in the
right state if it wasn't booted with redundancy on.

Could you help me to understand this?

> the host to passive BMC link may not be in the
right state if it wasn't booted with redundancy on.

Do you mean it could be kind of KCS channel which helps to trigger a failover?

I would like to know how do we handle below case.

1. Redundancy enabled and both BMC is in good state. 
2. Host is running.
3. Active BMC hit some issues which could trigger a failover.
4. Passive BMC changed into Active and Redundancy is disabled due the old Active BMC could not recover.
5. BMC CM operation is attempted to replace failed BMC.

Here, the Redundancy won't be enabled back because its locked since the Host is running?

If its true then Is it expected behaviour? I was under impression that we could enable redundancy back even if the host is running after the CM operation? is it not true? it could because we will play some cable or some other operation during the CM operation thats disturb host so we need to boot the host again after the CM operation?

redundant-bmc/src/services_impl.cpp
spinler: As we discussed, I think I can leave out the check for that since it will always move one of those other 2 states afterwards.

spinler:
> > This is because the host may treat the passive BMC differently if it
> > isn't enabled, and so the host to passive BMC link may not be in the
> > right state if it wasn't booted with redundancy on.
> 
> Could you help me to understand this?
> 
> > the host to passive BMC link may not be in the
> > right state if it wasn't booted with redundancy on.
> 

Ellen said that unless PHYP goes through its boot with redundancy enabled, the PCIe link between the passive BMC and the host might not be initialized.  Because it isn't initialized, if it were to become active the full PLDM interface wouldn't work.


> Do you mean it could be kind of KCS channel which helps to trigger a failover?

No, the KCS interface would still work, so a failover could be triggered.  It's the PCIe interface that is the issue.


> I would like to know how do we handle below case.
> 
>     1. Redundancy enabled and both BMC is in good state.
> 
>     2. Host is running.
> 
>     3. Active BMC hit some issues which could trigger a failover.
> 
>     4. Passive BMC changed into Active and Redundancy is disabled due the old Active BMC could not recover.
> 
>     5. BMC CM operation is attempted to replace failed BMC.


In that case:
- On the original active BMC it has the status that redundancy was enabled at runtime. (data::key::redundancyOffAtRuntime has valid=true,off=false)
- The passive BMC doesn't have that status since it isn't active and it isn't synced), but I don't think it needs it because it knows redundancy is already enabled so things must be fine. (need to think about that more, if the passive ever needs this from the active.)
- The passive BMC allows the failover, as the 'no redundancy at runtime' check isn't involved in allowing failovers, only when enabling redundancy. see next step.
- After the passive switches to active and the new passive comes back, it will see if it can re-enable redundancy.  Because data::key::redundancyOffAtRuntime was never written with (valid=true,off=true) on the passive, that check won't block redundancy.

I'll think more about the failover cases to see if there's a reason that info would need to be synced.

> Here, the Redundancy won't be enabled back because its locked since the Host is running?
> 
> If its true then Is it expected behaviour? I was under impression that we could enable redundancy back even if the host is running after the CM operation? is it not true? it could be because we will play some cable or some other operation during the CM operation thats disturb host so we need to boot the host again after the CM operation?

CM would be a new case that I didn't mention because I don't feel like we know the flows enough.  I envision some call that would come in where the code could clear data::key::redundancyOffAtRuntime  like it's doing when it detects a power off.  I will update the commit message and mention that.

spinler:
> I'll think more about the failover cases to see if there's a reason that info would need to be synced.

I can't think of a reason the passive BMC would need this info, because the only way it can become active at runtime is on a failover, and since the active BMC let redundancy become enabled then everything was fine.

It did make me realize that I should delete that info on the passive BMC when it starts, because theoretically it could have an old value from when it was active however long ago that said 'no redundancy at runtime'.

spinler:
> It did make me realize that I should delete that info on the passive BMC when it starts, 

The new push had just that fix.

RameshIyyar:
> it could have an old value from when it was active however long ago that said 'no redundancy at runtime' and then if it failed over at runtime it might read it and have the wrong value.

Right, because in the future, we might need to sync to retrieve the failover pause reason or other related details, which could potentially block us from re-enabling redundancy—such as in the CM flow.

> Ellen said that unless PHYP (and maybe hostboot?) goes through its boot with redundancy enabled, the PCIe link between the passive BMC and the host might not be initialized. Because it isn't initialized, if it were to become active the full PLDM interface wouldn't work.

Thanks for the detailed info.

From what I understand, during BSOC, we need to validate the PCI link—may not be PCI link training, but verify whether the link is good. If one of the BMCs has a bad PCI link, we should trigger a failover and disable redundancy. If both BMCs have bad PCI links, then redundancy should be disabled altogether.

I'm also unsure whether PCI link training occurs during the Hostboot IPL stage and whether it can fail at that point. If that's the case, we might need to support a failover or redundancy disable mechanism via a PLDM API. I’ll dig into it further and get back with more details.

RameshIyyar (approved):

spinler:
> From what I understand, during BSOC, we need to validate the PCI link—may not be PCI link training, but verify whether the link is good. If one of the BMCs has a bad PCI link, we should trigger a failover and disable redundancy. If both BMCs have bad PCI links, then redundancy should be disabled altogether.

> I'm also unsure whether PCI link training occurs during the Hostboot IPL stage and whether it can fail at that point. If that's the case, we might need to support a failover or redundancy disable mechanism via a PLDM API. I’ll dig into it further and get back with more details.

We'll see how it actually turns out, but my understanding is hostboot/bmc needs to train the PCIe link for the active BMC only, as that's what will be used for the hostboot PLDM communication, and it's up to PHYP to train the link of the passive BMC.  A fail training the active's link would trigger the BSOC flows.

RameshIyyar:
> > From what I understand, during BSOC, we need to validate the PCI link—may not be PCI link training, but verify whether the link is good. If one of the BMCs has a bad PCI link, we should trigger a failover and disable redundancy. If both BMCs have bad PCI links, then redundancy should be disabled altogether.
> 
> > I'm also unsure whether PCI link training occurs during the Hostboot IPL stage and whether it can fail at that point. If that's the case, we might need to support a failover or redundancy disable mechanism via a PLDM API. I’ll dig into it further and get back with more details.
> 
> We'll see how it actually turns out, but my understanding is hostboot/bmc needs to train the PCIe link for the active BMC only, as that's what will be used for the hostboot PLDM communication, and it's up to PHYP to train the link of the passive BMC. A fail training the active's link would trigger the BSOC flows.

Hmm, I think this involves many components.

BMC ISTEP: Involves PHAL to handle failover (if the other link is good) or to disable redundancy (if both links are bad).

SBE ISTEP: Involves PHAL and/or BMC PRD to handle failover or disable redundancy.

HBBL/HB/HBRT ISTEP: Could involve PHAL, BMC PRD, or possibly PLDM itself to manage failover or disable redundancy.

PHYP: The components involved depend on the type of error:

PHAL/HBRT: If the Active/Passive BMC has a bad link.

P-BMC: If only the Active BMC has a bad link.

PHYP: Might trigger a BMC reset/recover (R/R) if both BMC links are bad.

PHYP → HMC?: If PHYP can’t handle the issue when both BMC links are down. In that case, BMC R/R might help—but it’s unclear.

I think we need to discuss all of these flows to fully understand the impact on the RBMC manager and identify all relevant stakeholders. Initially, I thought only the BSOC was responsible for handling failover (if one link is good) or disabling redundancy (if both links are bad). But after thinking deeper, it seems that several other components might also be involved.

I mean the RBMC D-Bus API could be used by PHAL, HW-Diags, PLDM, or KCS. Depending on which component initiates the request based on the error and, certain states might need to be tracked in order to disable redundancy as per need based on the failure. I will capture this point in my Action list as well so that we won't miss it.

PR #83, opened 2025-04-14
RBMC: Create Providers class to hold the various provider helper classes

spinler:
Add a Providers class to manage the various provider helper classes, as there are more coming.  With this, only it has to be created in main(), and only it has to be passed around.

### DIFF
--- a/redundant-bmc/src/manager.cpp
+++ b/redundant-bmc/src/manager.cpp
@@ -14,10 +14,8 @@ namespace rbmc
 const std::chrono::minutes siblingTimeout{6};
 
 Manager::Manager(sdbusplus::async::context& ctx,
-                 std::unique_ptr<Services>&& services,
-                 std::unique_ptr<Sibling>&& sibling) :
-    ctx(ctx), redundancyInterface(ctx, *this), services(std::move(services)),
-    sibling(std::move(sibling))
+                 std::unique_ptr<Providers>&& providers) :
+    ctx(ctx), redundancyInterface(ctx, *this), providers(std::move(providers))
 {
     try
     {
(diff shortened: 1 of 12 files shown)


redundant-bmc/src/manager.cpp
```diff
@@ -68,17 +69,24 @@ sdbusplus::async::task<> Manager::startup()
 
     if (!passiveRoleInfo)
     {
-        if (sibling->isBMCPresent())
+        if (sibling.isBMCPresent())
         {
-            co_await sibling->waitForSiblingUp(siblingTimeout);
+            co_await sibling.waitForSiblingUp(siblingTimeout);
+
+            // Sibling service may have died.  Check again.
```
geissonator, line 76: This doesn't seem related to the commit?

spinler: looks like bad rebase, will fix.

spinler: pushed fix to just this.

geissonator (approved):

redundant-bmc/src/manager.cpp
```diff
@@ -53,8 +51,11 @@ Manager::Manager(sdbusplus::async::context& ctx,
 // NOLINTNEXTLINE
 sdbusplus::async::task<> Manager::startup()
 {
-    co_await sdbusplus::async::execution::when_all(services->init(),
-                                                   sibling->init());
+    auto& services = providers->getServices();
```
RameshIyyar, line 54: Can we not have an aggregation reference member for these providers?

no harm, just checking...Looks like we are creating a local reference in more places.

RameshIyyar (approved):

redundant-bmc/src/manager.cpp
spinler: I think for now I'd like to just pass around the Providers class, so that as new ones are added I don't have to change constructors.  I can revisit when they are all in.

PR #84, opened 2025-04-24
Set FailoversPaused when the system state is not off or runtime

spinler:
Add support for calculating FailoversPaused based on the system state.  The code is similar to the code that determines if redundancy can be enabled.

### DIFF
--- a/redundant-bmc/src/rbmc_tool.cpp
+++ b/redundant-bmc/src/rbmc_tool.cpp
@@ -39,6 +39,29 @@ sdbusplus::async::task<std::string> getBMCState(const rbmc::Services& services)
     }
 }
 
+void printNoRedReasons()
+{
+    using NoRedDetails =
+        std::map<rbmc::redundancy::NoRedundancyReason, std::string>;
+    auto details = data::read<NoRedDetails>(data::key::noRedDetails)
+                       .value_or(NoRedDetails{});
+    std::cout << std::format("Reasons for no BMC redundancy:\n");
+    if (!details.empty())
+    {
+        for (const auto& d : std::views::values(details))
+        {
+            std::cout << std::format("    {}\n", d);
+        }
+    }
+    else
+    {
+        // There can be long periods where the active BMC is waiting
+        // for the passive BMC so redundancy can't be checked yet.
+        // As far as rbmctool goes, label them as in a transition.
+        std::cout << std::format("    In transition\n");
+    }
+}
+
 // NOLINTBEGIN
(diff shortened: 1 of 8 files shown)


redundant-bmc/src/rbmc_tool.cpp
```diff
@@ -117,6 +136,11 @@ sdbusplus::async::task<> displayLocalBMCInfo(sdbusplus::async::context& ctx,
             {
                 printNoRedReasons();
             }
+
+            if ((role == "Active") && enabled && paused)
```
RameshIyyar, line 140: I believe we need to print the FO pause reason in the Passive BMC as well, but I don't think we've decided yet how it will obtain the FO pause reason.

redundant-bmc/src/redundancy_mgr.cpp
```diff
@@ -187,7 +199,7 @@ void RedundancyMgr::systemStateChange(SystemState newState)
 
     systemState = newState;
 
-    // TODO: on any change recalculate failovers paused
+    determineAndSetFailoversPaused();
```
RameshIyyar, line 202: https://github.com/ibm-openbmc/phosphor-state-manager/pull/84/commits/2cd9b6fb1f06f63a917a8f1e4b36decfa0450683#diff-bc7c6d2f949016fa3396798fab8c9c3c2ed18bbc298a323d6a01b4717ae11469L169 is it needed, now?

redundant-bmc/src/rbmc_tool.cpp
spinler: Yea, I think we'll want to do that when we have the data on that BMC.  I have stories opened to deal with the passive side of things.

redundant-bmc/src/redundancy_mgr.cpp
spinler: I can't quite follow what you mean here...  Can you elaborate?

RameshIyyar: Oh, it looks like the link to the untouched code isn't displaying correctly.

What I meant to say is:
In this function, I noticed the following snippet:

```
else if (newState == SystemState::booting)
{
    // TODO: failovers paused, etc
}

```

I think this block may be unnecessary now, since you're recalculating the failover points (FOP) anyway.

spinler: Thanks, yea I will remove that.

spinler: Fixed.

RameshIyyar (approved):

PR #85, opened 2025-05-02
RBMC: Passive BMC gets FO Paused from Active BMC

spinler:
The passive BMC will watch the sibling BMC's failovers paused property. If the sibling has the active role, mirror that value out to the FailoversPaused property on the main redundancy D-Bus interface.

This way, when the active BMC pauses failovers, the passive BMC will know that and will be able to reject any failover requests.

Right now, the sibling daemon is just mirroring the Redundancy.FailoversPaused property into the sibling interface, so the flow is:
1. Active BMC determines failovers must be paused.
2. Active BMC writes Redundancy.FailoversPaused.
3. Sibling daemon on active BMC sees that change and sends it to the passive BMC. (Writes to CFAM in IBM's case.)
4. Passive BMC sees sibling interface change and writes it to its Redundancy.FailoversPaused property. (This commit)

In the future, the passive BMC will have its own reasons for pausing failovers, so its Redundancy.FailoversPaused property would be the combination of the failovers paused value from the active BMC and its own calculated value.  And then the active BMC would do the same thing, combining its own calculated value with the one from the passive BMC.

Tested:

Failovers paused on active BMC:
```
Local BMC
-----------------------------
Role:                Active
BMC Position:        0
Redundancy Enabled:  true
BMC State:           Ready
Failovers Paused:    true
FW version hash:     87FC6A7D
Provisioned:         true
Reasons for failovers paused:
    System state is not off or runtime
```

Failovers paused now set on passive BMC's Redundancy iface:
```
$ busctl get-property xyz.openbmc_project.State.BMC.Redundancy \
/xyz/openbmc_project/state/bmc0 xyz.openbmc_project.State.BMC .Redundancy \
FailoversPaused
b true
```

Unpause failovers on active BMC:
```
Local BMC
-----------------------------
Role:                Active
BMC Position:        0
Redundancy Enabled:  true
BMC State:           Ready
Failovers Paused:    false
FW version hash:     87FC6A7D
Provisioned:         true
```

The passive BMC now shows FailoversPaused is false:
```
$ busctl get-property xyz.openbmc_project.State.BMC.Redundancy \
/xyz/openbmc_project/state/bmc0 xyz.openbmc_project.State.BMC .Redundancy \
FailoversPaused
b false
```

Change-Id: Ic956e7f52201cfcede0a81fa732cd8023cf4651c

### DIFF
--- a/redundant-bmc/src/passive_role_handler.cpp
+++ b/redundant-bmc/src/passive_role_handler.cpp
@@ -29,6 +29,8 @@ sdbusplus::async::task<> PassiveRoleHandler::start()
     // Setup the mirroring of the active BMC RedundancyEnabled
     setupSiblingRedEnabledWatch();
 
+    setupSiblingFailoversPausedWatch();
+
     try
     {
         // Only the active needs NoRedundancyDetails persisted.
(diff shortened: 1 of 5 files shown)


RameshIyyar (commented):
Is this PR intended to mirror the Active BMC FO pause property?
Are you planning to push a separate PR to mirror the Passive BMC FO pause property?

spinler:
> Is this PR intended to mirror the Active BMC FO pause property?

This PR is one way, active -> passive.

> Are you planning to push a separate PR to mirror the Passive BMC FO pause property?

Yea, though it isn't necessary until there is code on the passive BMC to determine if it wants FOs paused (checking for in progress code update).  I'm also going to have to change rbmc-cfamd to not just write Redundancy.FailoversPaused into the CFAM.

RameshIyyar (approved):

PR #86, opened 2025-05-08
Manage RBMC Data Sync on the active BMC

spinler:
Do a full sync when enabling redundancy, including any time the passive BMC's heartbeat starts back up.

Stop background syncing when redundancy is disabled.

Also watch the background sync status.  If it goes to critical, disable redundancy (assuming the heartbeat is still OK).

### DIFF
--- a/redundant-bmc/src/active_role_handler.cpp
+++ b/redundant-bmc/src/active_role_handler.cpp
@@ -36,11 +36,13 @@ sdbusplus::async::task<> ActiveRoleHandler::start()
             sibling.waitForSiblingRole(), sibling.waitForBMCSteadyState());
     }
 
-    redMgr.determineAndSetRedundancy();
+    co_await redMgr.determineRedundancyAndSync();
 
     // TODO: Create an error if no redundancy
 
     startSiblingWatches();
+
+    startSyncHealthWatch();
 
     co_return;
 }
@@ -92,11 +94,62 @@ sdbusplus::async::task<> ActiveRoleHandler::siblingHBStarted()
         sibling.waitForSiblingRole(), sibling.waitForBMCSteadyState());
 
     lg2::info("Attempting to enable redundancy now that sibling is back");
-    redMgr.determineAndSetRedundancy();
+    co_await redMgr.determineRedundancyAndSync();
 
     // TODO: full sync, etc
 
     startSiblingWatches();
+
+    co_return;
+}
+
+void ActiveRoleHandler::syncHealthPropertyChanged(
+    SyncBMCData::SyncEventsHealth health)
+{
+    lg2::info("Sync health property changed to {HEALTH}", "HEALTH", health);
+
+    // Don't care about changes if no redundancy.
+    if (!redundancyInterface.redundancy_enabled())
+    {
+        return;
+    }
+
+    if (health == SyncBMCData::SyncEventsHealth::Critical)
+    {
+        ctx.spawn(syncHealthCritical());
+    }
+}
+
+// NOLINTNEXTLINE
+sdbusplus::async::task<> ActiveRoleHandler::syncHealthCritical()
+{
+    using namespace std::chrono_literals;
+
+    lg2::info("Disabling background sync because it is failing");
+    co_await providers.getSyncInterface().disableBackgroundSync();
+
+    // A passive BMC reboot should not result in redundancy being
+    // disabled, so wait a bit for the passive BMC's heartbeat
+    // to change.  If it's still running, then this is a valid
+    // sync fail so disable redundancy.  If it isn't running then
+    // then the code that deals with the sibling heartbeat will
+    // deal with it.
+    // TODO: Also do network failure detection.
+    lg2::info("Waiting to see if sibling heartbeat stops");
+    co_await providers.getSibling().pauseForHeartbeatChange();
+
+    if (providers.getSibling().hasHeartbeat())
+    {
+        lg2::error("Disabling redundancy due to critical sync health");
+
+        // This will disable redundancy
+        redMgr.handleBackgroundSyncFailed();
+    }
+    else
+    {
+        lg2::info(
+            "Sync health is critical, but there is also a sibling heartbeat loss");
+    }
 
     co_return;
 }
--- a/redundant-bmc/src/redundancy_mgr.cpp
+++ b/redundant-bmc/src/redundancy_mgr.cpp
@@ -64,7 +64,8 @@ redundancy::NoRedundancyReasons RedundancyMgr::getNoRedundancyReasons()
         .codeVersionsMatch =
             services.getFWVersion() == sibling.getFWVersion().value_or(""),
         .manualDisable = manualDisable,
-        .redundancyOffAtRuntimeStart = isRedundancyOffAtRuntime()};
+        .redundancyOffAtRuntimeStart = isRedundancyOffAtRuntime(),
+        .syncFailed = syncFailed};
 
     auto reasons = redundancy::getNoRedundancyReasons(input);
 
--- a/redundant-bmc/src/redundancy_mgr.cpp
+++ b/redundant-bmc/src/redundancy_mgr.cpp
@@ -46,6 +46,50 @@ void RedundancyMgr::determineAndSetRedundancy()
     redundancyDetermined = true;
 
     determineAndSetFailoversPaused();
+
+    if (!redundancyInterface.redundancy_enabled())
+    {
+        // Make sure syncs are disabled if redundancy is disabled.
+        ctx.spawn(providers.getSyncInterface().disableBackgroundSync());
+    }
+}
+
+// NOLINTNEXTLINE
+sdbusplus::async::task<> RedundancyMgr::determineRedundancyAndSync()
+{
+    syncFailed = false;
+    determineAndSetRedundancy();
+
+    if (redundancyInterface.redundancy_enabled())
+    {
+        try
+        {
+            if (!co_await providers.getSyncInterface().doFullSync())
+            {
+                lg2::error("Disabling redundancy because full sync failed");
+                syncFailed = true;
+            }
+        }
+        catch (const std::exception& e)
+        {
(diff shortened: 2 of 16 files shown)


redundant-bmc/src/sync_interface_impl.cpp
```diff
@@ -0,0 +1,135 @@
...
+        {
+            // A full sync is already running, don't need to start one.
+            lg2::info(
+                "A full sync is already in progress, waiting for completion");
+        }
+
+        do
+        {
+            using namespace std::chrono_literals;
+            co_await sdbusplus::async::sleep_for(ctx, 1s);
```
RameshIyyar: Is it really necessary to check at 1-second intervals? Currently, it's taking 35 seconds, so it would be more efficient to use the property change signal instead?

redundant-bmc/src/active_role_handler.cpp
```diff
@@ -92,7 +94,7 @@ sdbusplus::async::task<> ActiveRoleHandler::siblingHBStarted()
         sibling.waitForSiblingRole(), sibling.waitForBMCSteadyState());
 
     lg2::info("Attempting to enable redundancy now that sibling is back");
-    redMgr.determineAndSetRedundancy();
+    co_await redMgr.determineRedundancyAndSync();
```
RameshIyyar, line 97: If the Active BMC reboots, there’s a possibility that PELs might have been created, deleted, or updated during that time in the Passive BMC. So, we’ll need to perform a full sync from Passive to Active in such cases as well.

Are you planning to handle that in a separate PR?

redundant-bmc/src/sync_interface_impl.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+// NOLINTNEXTLINE
+sdbusplus::async::task<bool> SyncInterfaceImpl::doFullSync()
+{
+    using SyncStatus = SyncBMCData::FullSyncStatus;
+    SyncStatus status = SyncStatus::Unknown;
+
+    try
+    {
+        co_await lookupService();
```
RameshIyyar, line 22: Do you prefer to handle the data sync restart scenario within the RBMC manager?

Currently, whenever data sync starts or restarts, it performs a full sync followed by background sync if Redundancy is enabled.

(We still need to add code to ensure background sync is not started when Redundancy is disabled during the data sync starts)

redundant-bmc/src/active_role_handler.cpp
```diff
@@ -101,4 +103,55 @@ sdbusplus::async::task<> ActiveRoleHandler::siblingHBStarted()
...
+
+// NOLINTNEXTLINE
+sdbusplus::async::task<> ActiveRoleHandler::syncHealthCritical()
+{
+    using namespace std::chrono_literals;
+
+    lg2::info("Disabling background sync because it is failing");
+    co_await providers.getSyncInterface().disableBackgroundSync();
+
+    // A passive BMC reboot should not result in redundancy being
```
RameshIyyar, line 131: If the data sync, stunnel, or rsync services fail to restart (despite systemd’s restart policy), the BMC should transition to a "Quiesced" state, allowing HMC/PHYP to trigger a failover.

I'm trying to think how we should handle the scenario where one side fails to start these services, causing sync from the other side to fail. This would effectively disable redundancy, with the failure reason captured as "sync failure" along with a PEL containing the appropriate callouts (including a procedure callout if the root cause is a firmware bug). In this case, replacing the BMC is not necessary. If it’s determined to be a firmware issue, a concurrent code update (CU) should be sufficient to resolve it.

So the question is: in such situations, should we treat redundancy as lost until a firmware update is applied? And is it acceptable to enforce this rule for firmware-related failures, or should we allow failover using the existing data, even with the known issue present until the FW update?

redundant-bmc/src/redundancy_mgr.cpp
```diff
@@ -46,6 +46,50 @@ void RedundancyMgr::determineAndSetRedundancy()
...
+sdbusplus::async::task<> RedundancyMgr::determineRedundancyAndSync()
+{
+    syncFailed = false;
+    determineAndSetRedundancy();
+
+    if (redundancyInterface.redundancy_enabled())
+    {
+        try
+        {
+            if (!co_await providers.getSyncInterface().doFullSync())
```
RameshIyyar, line 67: We’ve discussed this before but haven’t yet concluded how to handle the following case:

We plan to delay the full sync until all FRU VPD collection is either completed or has failed. This should be handled automatically, as the BMC will remain in the "Ready" state (I hope we are considering local BMC state as well) both only after VPD collection finishes.

However, there’s still a possibility that the BMC could enter a "Quiesced" state due to an other error. If a failover is triggered in that state, we risk syncing partial VPD data.

To prevent this, we should ensure that neither full sync nor background sync begins until the VPD collection is fully completed or we need to reboot the Passive BMC during the FO in worst case like you mentioned.

redundant-bmc/src/active_role_handler.cpp
```diff
@@ -101,4 +103,55 @@ sdbusplus::async::task<> ActiveRoleHandler::siblingHBStarted()
...
+    // disabled, so wait a bit for the passive BMC's heartbeat
+    // to change.  If it's still running, then this is a valid
+    // sync fail so disable redundancy.  If it isn't running then
+    // then the code that deals with the sibling heartbeat will
+    // deal with it.
+    // TODO: Also do network failure detection.
+    lg2::info("Waiting 5s to see if sibling heartbeat stops");
+    co_await sdbusplus::async::sleep_for(ctx, 5s);
```
geissonator: Is this 5s hard coded somewhere else as well in the code that handles the heartbeat? Seems dangerous to have it hard coded in multiple places.

redundant-bmc/src/sync_interface_impl.cpp
spinler: That should be fine to change.  It would mean we couldn't ever check for it to time out, but that should be fine I would think.

redundant-bmc/src/active_role_handler.cpp
spinler: Yea, I have a different story for that.  I've had good path working for a while now in testing, where it will start the full sync when RedundancyEnabled from the active BMC changes to true.  I'm not sure yet on the error paths, if it fails what needs to be done if anything.

redundant-bmc/src/sync_interface_impl.cpp
spinler: I'd rather avoid having to watch for the sync daemon crashing/restarting, so I think what is being done now is fine right?

redundant-bmc/src/active_role_handler.cpp
spinler: I can move it into function.

redundant-bmc/src/redundancy_mgr.cpp
spinler: As far as holding off enabling redundancy until after VPD collection, my latest thought was maybe we could make a service that links into obmc-bmc-active.target.wants that waits until VPD collection is complete.  This avoids the problem of a BMC going to Quiesced while VPD was still in progress, which would move into enabling redundancy, without my code having to make an IBM specific call to check VPD-complete D-Bus.

then it would look like:
1. rbmc-mgr determines itself to be active
2. rbmc-manager starts obmc-bmc-active.target
    1. VPD Collection starts
    2. wait-for-vpd.service (new) starts 
    3. Is there still an old wait-for-vpd.service running here to hold off BMC ready?
4. While VPD collection continues:
    1. obmc-active.target not active yet, rbmc-mgr is waiting inside Services::startUnit() for state == "active"
5. VPD collection completes
    1. wait-for-vpd.service completes so obmc-bmc-active.target completes
    2. startUnit() completes
6. rbmc-mgr attempts to enable redundancy and starts full sync if it can.

redundant-bmc/src/active_role_handler.cpp
spinler: That sounds like a pretty rare condition that would hopefully never happen.  But yea, I don't know.

redundant-bmc/src/sync_interface_impl.cpp
spinler: Done.

redundant-bmc/src/active_role_handler.cpp
spinler: Done.

geissonator (approved):
I'm good if Ramesh is.

redundant-bmc/src/sync_interface_impl.cpp
RameshIyyar: Yes, that’s fine. I was just wondering whether all full sync scenarios need to be handled exclusively by the RBMC manager.

redundant-bmc/src/active_role_handler.cpp
RameshIyyar: Hmm its more like Redundancy policy specific thing to decide.

redundant-bmc/src/redundancy_mgr.cpp
RameshIyyar: > As far as holding off enabling redundancy until after VPD collection

Hmm that would be ideal solution.

> we could make a service that links into obmc-bmc-active.target.wants that waits until VPD collection is complete. This avoids the problem of a BMC going to Quiesced while VPD was still in progress, which would move into enabling redundancy, without my code having to make an IBM specific call to check VPD-complete D-Bus.

Agree, IBM business logic need to sit outside of the RBMC manager.

>> rbmc-manager starts obmc-bmc-active.target
> VPD Collection starts
> wait-for-vpd.service (new) starts
> Is there still an old wait-for-vpd.service running here to hold off BMC ready?

Hmm I am not sure whether we could able to use the existing `wait-vpd-parsers` service to hold BMC state (Ready/Quiesced) update. @geissonator Any thoughts?

Anyway, this is a separate item to address, and since the VPD work is still in progress, we can track it separately.

RameshIyyar (approved):

PR #87, opened 2025-05-12
obmcutil: check for transition states on commands

geissonator:
When obmcutil was originally written, the design was that users know what they are doing so it would allow them to shoot themselves in the foot if they wanted. It was a lab/debug tool. Proper external interfaces like the Web UI enforce when power on and off operations can be issued.

As time has gone on, the use of OpenBMC has increased significantly and so has the userbase of obmcutil. Within IBM we have dozens of different groups interacting with OpenBMC servers and hundreds of users of obmcutil. The figuratively shooting of ones foot has become a problem.

We see multiple defects per month where the end result is that someone was using obmcutil in a way they should not. A lot of times the bad behavior is buried in some automated script that the person opening the defect doesn't even know about. It results in countless hours of developer debug to determine root cause.

This commit will not fix all of the issues we've seen but it will fix a lot of them by ensuring a user does not issue a obmcutil power on/off command when one is already in process.

Tested:
- Set Chassis state to TransitioningToOff and confirmed a power on/off was not allowed
- Set Host state to .TransitioningToOff and confirmed a power on/off was not allowed
- Did above tests with TransitioningToOn/TransitioningToRunning
- Verified good path poweron
- Verified --force allows a user to override the new check

Change-Id: Ie7123f37dd98ffd388f79ff7b31662c30a730bcf

### DIFF
--- a/obmcutil
+++ b/obmcutil
@@ -20,8 +20,7 @@ CONTROL_OBJECT=$OBJECT_ROOT/control
 HOST_TIMEOUT_TARGET=obmc-host-timeout@0.target
 HOST_CRASH_TARGET=obmc-host-crash@0.target
 
-## NOTE: The following global variables are used only in the run_timeout cmd.
-## By declaring these globally instead of passing them through the
+## NOTE: By declaring these globally instead of passing them through the
 ## intermediary functions, which may not be "best practice", the readability
 ## and cleanliness of the code should at least be increased.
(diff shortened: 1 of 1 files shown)


geissonator:
Upstream: https://gerrit.openbmc.org/c/openbmc/phosphor-state-manager/+/79779

PR #89, opened 2025-06-03
bmc-state: ensure bmc remains in quiesced

geissonator:
Recently found a bug where if a critical application fails early in the BMC boot, the BMC state will start at Quiesced but go to Ready. This is a bug in that once the BMC goes to Quiesced, it must remain there until the BMC is rebooted.

The bug is that the discoverInitialState() function initially moves to Quiesced on startup but it does not unsubscribe from systemd events so if the failing service was not a "Requires" for multi-user.target, the BMC state management application will get the "done" dbus event and move from Quiesced to Ready.

Tested:
- Confirmed that BMC now remains in Quiesced when early services fail

Change-Id: Ied1a25939b324c04ae850894584daabe6f02c190


geissonator:
Upstream: https://gerrit.openbmc.org/c/openbmc/phosphor-state-manager/+/80757

PR #92, opened 2025-06-03
RBMC: Create active-bmc-only.conf file

anoo1:


service_files/active-bmc-only.conf
```diff
@@ -0,0 +1,4 @@
+[Unit]
+ConditionPathExists=/run/openbmc/active-bmc
+After=phosphor-broadcast-active-bmc-role.service
```
spinler, line 3: I don't think we want this After= line.  In other places you don't usually see the After= inside a service that's already a dependency on that thing.

spinler: nevermind.

service_files/phosphor-broadcast-active-bmc-role.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Broadcast if the BMC role is Active
+PartOf=obmc-bmc-active.target
+
```
anoo1, line 4: Testing with Matt, the After in this file is the one that needed to be removed, because the conf file says it's part of the active target, and runs after this broadcast service, so this service can't say to run after the active target.

spinler (approved):

geissonator (commented):
I think this part of the commit message can be removed?
```
The 'DefaultDependencies=no' is required so that there aren't error
messages like:
...
```

service_files/phosphor-broadcast-active-bmc-role.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Broadcast if the BMC role is Active
```
geissonator: The "if" part confuses me a bit here. If this service runs, then the BMC "is" the active right? So service should really be "Broadcast that this BMC role is Active'

anoo1: Yeah agree. Updated.

anoo1:
> I think this part of the commit message can be removed?

Ah yes, done.

geissonator (approved):

PR #93, opened 2025-06-06
RBMC: Stop using 2 sibling BMC properties

spinler:
Stop using the sibling BMC position and the sibling 'sibling Communication OK' properties.
This is to prepare for moving away from the 'xyz.openbmc_project.State.BMC.Redundancy.Sibling' interface.

These properties don't have a good new home, and turned out not to be that necessary anyway.  Details are in the commit messages.

### DIFF
--- a/redundant-bmc/src/rbmc_tool.cpp
+++ b/redundant-bmc/src/rbmc_tool.cpp
@@ -193,14 +193,12 @@ sdbusplus::async::task<> displaySiblingBMCInfo(sdbusplus::async::context& ctx,
         auto bmcState = BMCState::convertBMCStateToString(
             std::get<BMCState::BMCState>(props.at("BMCState")));
         bmcState = bmcState.substr(bmcState.find_last_of('.') + 1);
-        auto pos = std::get<size_t>(props.at("BMCPosition"));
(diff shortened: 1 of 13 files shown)


geissonator (commented):
nit: "which which" in commit message.

redundant-bmc/src/rbmc_tool.cpp
```diff
@@ -193,14 +193,12 @@ sdbusplus::async::task<> displaySiblingBMCInfo(sdbusplus::async::context& ctx,
         auto bmcState = BMCState::convertBMCStateToString(
             std::get<BMCState::BMCState>(props.at("BMCState")));
         bmcState = bmcState.substr(bmcState.find_last_of('.') + 1);
-        auto pos = std::get<size_t>(props.at("BMCPosition"));
```
geissonator, line 196: Why position removal in the "comm ok" commit?

spinler: good catch, will move to other one.

RameshIyyar (approved):

spinler:
> nit: "which which" in commit message.

Thanks, fixed.

PR #96, opened 2025-06-13
RBMC: Start using the 4 new Sibling D-bus interfaces

spinler:
The rbmc-cfamd daemon now uses the 4 different D-Bus interfaces to hold the sibling BMC's information instead of just 1.  Move over the phosphor-rbmc-state-manager daemon to start using them.

### DIFF
--- a/redundant-bmc/src/sibling_impl.cpp
+++ b/redundant-bmc/src/sibling_impl.cpp
@@ -3,11 +3,26 @@
 
 #include <phosphor-logging/lg2.hpp>
 #include <xyz/openbmc_project/ObjectMapper/client.hpp>
+#include <xyz/openbmc_project/Software/Version/common.hpp>
+#include <xyz/openbmc_project/State/BMC/Redundancy/common.hpp>
+#include <xyz/openbmc_project/State/BMC/common.hpp>
+#include <xyz/openbmc_project/State/Decorator/Heartbeat/common.hpp>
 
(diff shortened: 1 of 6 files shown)


geissonator (approved):

redundant-bmc/src/sibling_impl.cpp
```diff
@@ -297,11 +288,56 @@ sdbusplus::async::task<> SiblingImpl::watchNameOwnerChanged()
...
+    auto sibling = sdbusplus::async::proxy()
+                       .service(serviceName)
+                       .path(RedIntf::namespace_path::value)
+                       .interface("org.freedesktop.DBus.ObjectManager");
+    try
+    {
+        auto objects =
+            co_await sibling.call<ManagedObjects>(ctx, "GetManagedObjects");
+
+        for (const auto& interfaces : std::views::values(objects))
```
RameshIyyar: Why do we need to get from all objects? just getting from the SiblingBMC object is good enough, right? 

I think we will always get single object that would be bmc1, right? if so, cannot we just use the SiblingBMC segment?

spinler: Good point, I will change it.

spinler: Fixed.  Only change in new push.

RameshIyyar (approved):

PR #99, opened 2025-06-17
RBMC: Only allow failovers after full sync

spinler:
After the active BMC determined redundancy could be enabled, it was setting the 'RedundancyEnabled' and 'FailoversAllowed' properties to true and then starting a full sync.

However, it is not desired to actually allow a failover to occur until the full sync is complete.  So, add a new check into the code that calculates failovers allowed to also check for the full sync being complete.

The flag that tracks if the full sync is complete lasts until redundancy is disabled.

Tested:
Traces during initial full sync that show failovers not being allowed:
```
phosphor-rbmc-state-manager[1516]: Finished waiting for obmc-bmc-active.target to start (result = active)
phosphor-rbmc-state-manager[1516]: Done waiting for sibling steady state. State = xyz.openbmc_project.State.BMC.BMCState.Ready
phosphor-rbmc-state-manager[1516]: Enabling redundancy
phosphor-rbmc-state-manager[1516]: Failovers not allowed because A full sync hasn't been completed
phosphor-rbmc-state-manager[1516]: Starting full sync and waiting for completion
phosphor-rbmc-state-manager[1516]: Full sync completed with status xyz.openbmc_project.Control.SyncBMCData.FullSyncStatus.FullSyncCompleted
phosphor-rbmc-state-manager[1516]: Changing failovers to allowed
```

rbmctool output during that time:
```
Local BMC
-----------------------------
Role:                Active
BMC Position:        0
Redundancy Enabled:  true
BMC State:           Ready
Failovers Allowed:   false
FW version hash:     7969307F
Provisioned:         true
Reasons failovers are not allowed:
    A full sync hasn't been completed
```

Change-Id: I14dedd43619fffdeea11449319b681f8b95a1698

### DIFF
--- a/redundant-bmc/src/sync_interface_impl.cpp
+++ b/redundant-bmc/src/sync_interface_impl.cpp
@@ -19,6 +19,8 @@ sdbusplus::async::task<bool> SyncInterfaceImpl::doFullSync()
 
     try
     {
+        fullSyncComplete = false;
+
         co_await lookupService();
 
         auto sync = SyncBMCData(ctx)
@@ -74,7 +76,8 @@ sdbusplus::async::task<bool> SyncInterfaceImpl::doFullSync()
     lg2::info("Full sync completed with status {STATUS}", "STATUS", status);
 
     fullSyncInProgress = false;
-    co_return status == SyncStatus::FullSyncCompleted;
+    fullSyncComplete = status == SyncStatus::FullSyncCompleted;
+    co_return fullSyncComplete;
 }
 
 // NOLINTNEXTLINE
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ state to/by the end user.
 
 ## State Tracking and Control
 
-phosphor-state-manager makes extensive use of systemd. There is a writeup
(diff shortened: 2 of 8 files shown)


redundant-bmc/src/sync_interface_impl.cpp
```diff
@@ -74,7 +76,8 @@ sdbusplus::async::task<bool> SyncInterfaceImpl::doFullSync()
     lg2::info("Full sync completed with status {STATUS}", "STATUS", status);
 
     fullSyncInProgress = false;
-    co_return status == SyncStatus::FullSyncCompleted;
+    fullSyncComplete = status == SyncStatus::FullSyncCompleted;
```
RameshIyyar, line 79: Do we need to update flag if any exception as well otherwise we won't able to do FO, right?

spinler: At the beginning of the function on line 22 it gets set to false so that we don't need to handle that path.

RameshIyyar (approved):

geissonator (approved):

PR #100, opened 2025-06-26
Put failover interface on D-Bus

spinler:
Put the failover interface on D-Bus and add rbmctool options to call it.   The call doesn't do anything yet and just returns an 'Unavailable' error.

### DIFF
--- a/redundant-bmc/src/manager.cpp
+++ b/redundant-bmc/src/manager.cpp
@@ -11,8 +11,14 @@
 namespace rbmc
 {
 
+const std::string failoverPath =
+    std::string{RedundancyInterface::namespace_path::value} + '/' +
+    RedundancyInterface::namespace_path::bmc;
+
 Manager::Manager(sdbusplus::async::context& ctx,
                  std::unique_ptr<Providers>&& providers) :
+    sdbusplus::aserver::xyz::openbmc_project::control::Failover<Manager>(
+        ctx, failoverPath.c_str()),
     ctx(ctx), redundancyInterface(ctx, *this), providers(std::move(providers))
 {
     try
@@ -41,6 +47,9 @@ Manager::Manager(sdbusplus::async::context& ctx,
         lg2::error("Failed trying to obtain previous role error: {ERROR}",
                    "ERROR", e);
     }
+
+    // emit the Failover interfaces added signal
+    emit_added();
 
     ctx.spawn(startup());
 }
@@ -265,6 +274,15 @@ void Manager::disableRedPropChanged(bool disable)
     }
 
     handler->disableRedPropChanged(disable);
+}
+
+// NOLINTNEXTLINE
+sdbusplus::async::task<> Manager::method_call(
+    start_failover_t /* unused */, const FailoverOptions& /* options */)
+{
+    // TODO: Implement the failover
+    lg2::error("Failovers are not implemented yet");
+    throw sdbusplus::xyz::openbmc_project::Common::Error::Unavailable();
 }
 
 } // namespace rbmc
--- a/redundant-bmc/src/rbmc_tool.cpp
+++ b/redundant-bmc/src/rbmc_tool.cpp
@@ -6,7 +6,7 @@
 #include "sibling_reset_impl.hpp"
 
 #include <CLI/CLI.hpp>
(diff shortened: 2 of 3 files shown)


redundant-bmc/src/manager.cpp
```diff
@@ -11,8 +11,14 @@
 namespace rbmc
 {
 
+const std::string failoverPath =
+    std::string{RedundancyInterface::namespace_path::value} + '/' +
```
RameshIyyar, line 15: nit: cannot we just use instance path to get the BMC0 path?

RameshIyyar (approved):

redundant-bmc/src/manager.cpp
spinler: That isn't upstream so I removed all uses of it:
```
paths:
    - namespace: /xyz/openbmc_project/state
      segments:
          - name: BMC
            description: >
                The object representing _this_ BMC is always at bmc0.
            value: bmc0

```

RameshIyyar: oh yeah...I referred wrong branch.

RameshIyyar (approved):

redundant-bmc/src/rbmc_tool.cpp
```diff
@@ -295,6 +296,52 @@ sdbusplus::async::task<> modifyRedundancyOverride(
     }
 }
 
+// NOLINTNEXTLINE
+sdbusplus::async::task<> startFailover(sdbusplus::async::context& ctx,
+                                       bool force)
+{
+    std::map<std::string, std::variant<bool>> options;
```
geissonator: nit: I always am a bit unsettled when we are defining something for an interface in two different places. Could we at least put a "using FailoverOptions = std::map<std::string, std::variant<bool>>;" in here so a grep would fine the 2 definitions?

redundant-bmc/src/rbmc_tool.cpp
```diff
@@ -326,6 +375,14 @@ int main(int argc, char** argv)
     resetGroup->add_flag("--reset-sibling", resetSibling,
                          "Reset the sibling BMC");
 
+    auto* failoverGroup = app.add_option_group("Starting failovers");
+    auto* fo =
+        failoverGroup->add_flag("--failover", failover, "Start a failover");
+    failoverGroup
+        ->add_flag("--force-failover", forceFailover,
```
geissonator: Could we have a "-f" option as well to shorten things a bit? I could see this being used a lot in the lab for testing and such.

redundant-bmc/src/rbmc_tool.cpp
```diff
@@ -295,6 +296,52 @@ sdbusplus::async::task<> modifyRedundancyOverride(
...
+            .service(Redundancy::interface)
+            .path(path.str)
+            .start_failover(options);
+    }
+    catch (const sdbusplus::exception_t& e)
+    {
+        if (std::string{"xyz.openbmc_project.Common.Error.Unavailable"} ==
+            e.name())
+        {
+            std::cout
```
geissonator, line 334: lg2::error?

redundant-bmc/src/rbmc_tool.cpp
spinler: done.  I couldn't find a good header to just have it in one place.

redundant-bmc/src/rbmc_tool.cpp
spinler: There are 2 reasons why I didn't do it that way:
1. Since this is a terminal app, when lg2 writes it to the terminal, it adds the priority like <6> to the message, which always looks weird to me and I want to minimize.
2. I don't think this actually needs to be in the journal.  There is already an indication that the failover came from rbmctool lines 309/316, and the errors will always already be in the journal (which is why this message says to check the journal.

redundant-bmc/src/rbmc_tool.cpp
spinler: Yea I can add short options for these.  Initially I wasn't keen on that because I didn't want people to accidentally trigger a failover.

spinler: done.

PR #101, opened 2025-07-11
Check if a failover request can proceed

spinler:
In the StartFailover method, add in the code to check if it should be allowed to proceed.

### DIFF
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -188,7 +188,16 @@ restarted next time a full sync is done. It will do the same 5 second wait the
 active BMC does to see if the heartbeat has stopped, meaning most likely the
 active BMC either died or was rebooted.
 
-## Allowing Failovers
+## Failovers
+
+A failover is when redundancy is enabled and the passive BMC takes over as the
+active BMC, and the original active BMC becomes passive. Redundancy may or may
+not be re-enabled afterwards, depending on if something is preventing it or not.
+
+The failover is always driven by the original passive BMC, which then takes over
+as active.
+
+### Allowing Failovers
 
 Even when redundancy is enabled, there are periods when failovers will not be
 allowed. The `FailoversAllowed` D-Bus project reflects this state.
@@ -209,3 +218,18 @@ Future work to be done:
 - Determine if the other BMC needs the reasons, or just if FOs aren't allowed.
 - When writing the failover code, reject the failover if it isn't allowed,
   though there still needs to be a method to force it for use by field support.
+
+### Rejecting a failover request
+
+When the call is made to start the failover on the passive BMC, it will reject
+the request if any of the following are true.
+
+1. Redundancy isn't enabled.
+1. FailoversAllowed is false. Exceptions are:
+   - The `force` option was passed into the `StartFailover` method.
+   - The active BMC is in the `Quiesced` state.
+1. A full sync on the passive BMC is in progress.
+1. The active BMC has no heartbeat and redundancy wasn't last known to be
+   enabled. If it was last known to be enabled, a failover is allowed so that
+   the remaining BMC can become active.
+1. The passive BMC is not in the `Ready` state.
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -224,6 +224,7 @@ Future work to be done:
 When the call is made to start the failover on the passive BMC, it will reject
 the request if any of the following are true.
 
+1. A failover is imminent or already in progress.
 1. Redundancy isn't enabled.
 1. FailoversAllowed is false. Exceptions are:
    - The `force` option was passed into the `StartFailover` method.
--- a/redundant-bmc/src/manager.cpp
+++ b/redundant-bmc/src/manager.cpp
@@ -277,9 +277,31 @@ void Manager::disableRedPropChanged(bool disable)
 }
 
 // NOLINTNEXTLINE
-sdbusplus::async::task<> Manager::method_call(
-    start_failover_t /* unused */, const FailoverOptions& /* options */)
+sdbusplus::async::task<> Manager::method_call(start_failover_t /* unused */,
+                                              const FailoverOptions& options)
 {
+    if (!handler)
+    {
+        lg2::error("Failover not allowed because it is too early");
+        throw sdbusplus::xyz::openbmc_project::Common::Error::Unavailable();
+    }
+
+    if (redundancyInterface.failover_imminent() || failover_in_progress())
+    {
+        lg2::error(
+            "Failover not allowed because a failover is already imminent or in progress ");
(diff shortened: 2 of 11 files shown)


redundant-bmc/src/redundancy.cpp
```diff
@@ -177,4 +179,109 @@ std::string getFailoversNotAllowedDescription(FailoversNotAllowedReason reason)
...
+                // failovers-not-allowed so don't block it, just trace it.
+                lg2::info(
+                    "The sibling BMC is quiesced while failovers are not allowed");
+            }
+            else
+            {
+                return Reason::failoversNotAllowed;
+            }
+        }
+        else if (input.syncInProgress)
```
geissonator, line 217: Do we have situations where a bad primary BMC could cause this to be permanently set? Is there a timeout? This one worries me the most because it could occur for a variety of reasons and it would be fairly unexpected for the HMC or PHYP to get this response. Do we tell them to retry? In 30s? 1 minute?

redundant-bmc/test/redundancy_test.cpp
```diff
@@ -202,3 +202,103 @@ TEST(RedundancyTest, GetFailoversNotAllowedDescTest)
...
+    // Failovers not allowed, but sibling is Quiesced
+    {
+        auto input = golden;
+        input.failoversNotAllowed = true;
+        input.siblingState = rbmc::BMCState::Quiesced;
+        EXPECT_EQ(rbmc::fo_blocked::getFailoverBlockedReason(input),
+                  rbmc::fo_blocked::Reason::none);
+    }
+
+    // Redundancy not enabled
```
geissonator: Synch in progress

redundant-bmc/src/redundancy.cpp
spinler: @RameshIyyar Do you have any worries here about a full sync never completing?  How my doFullSync() code is written now it waits on a propertiesChanged signal so it can't time out.

RameshIyyar: It will eventually complete—either successfully or with a failure—but the time it takes is uncertain due to our retry mechanism.

RameshIyyar (commented):
Sorry I could not get chance to review this PR but just replied to the comments. I will review shortly.

redundant-bmc/test/redundancy_test.cpp
spinler: Thanks. Done.

redundant-bmc/README.md
```diff
@@ -209,3 +218,18 @@ Future work to be done:
...
+### Rejecting a failover request
+
+When the call is made to start the failover on the passive BMC, it will reject
+the request if any of the following are true.
+
+1. Redundancy isn't enabled.
+1. FailoversAllowed is false. Exceptions are:
+   - The `force` option was passed into the `StartFailover` method.
+   - The active BMC is in the `Quiesced` state.
+1. A full sync on the passive BMC is in progress.
```
RameshIyyar, line 232: Should we also consider a full sync from the Active BMC when the Passive BMC is handling the failover request?

Also, I believe there are a few more scenarios to consider, such as when a code update is in progress or VPD collection is underway.

redundant-bmc/src/redundancy.cpp
RameshIyyar: While thinking further, I was wondering—can we trigger some kind of event to HMC/PHYP to retry the failover, instead of relying on a timeout? If the timeout mechanism is straightforward/simple, that’s fine. I just want to confirm whether there’s any scenario where they might stop retrying altogether.


> Do we have situations where a bad primary BMC could cause this to be permanently set?

Also, I’d like to clarify @geissonator’s question — here we are checking the Full Sync status of the Passive BMC, not the Active one.

redundant-bmc/src/manager.cpp
```diff
@@ -277,9 +277,31 @@ void Manager::disableRedPropChanged(bool disable)
...
+    {
+        lg2::error("Failover not allowed because it is too early");
+        throw sdbusplus::xyz::openbmc_project::Common::Error::Unavailable();
+    }
+
+    if (redundancyInterface.failover_imminent() || failover_in_progress())
+    {
+        lg2::error(
+            "Failover not allowed because a failover is already imminent or in progress ");
+        throw sdbusplus::xyz::openbmc_project::Common::Error::Unavailable();
```
RameshIyyar, line 293: I think we should define a separate error for this or we can revisit it later if the DBus caller needs to handle the error differently.

redundant-bmc/README.md
spinler: > Should we also consider a full sync from the Active BMC when the Passive BMC is handling the failover request?

That one is covered by FailoversAllowed already.

> Also, I believe there are a few more scenarios to consider, such as when a code update is in progress 

Yea, when that code is done I'll add it here.

> or VPD collection is underway.

VPD collection will be occurring before redundancy is enabled.

redundant-bmc/src/manager.cpp
spinler: We could probably do different errors if necessary, though I'd rather get the Failover interface merged upstream first so any discussions about them don't hold up just getting it in.

redundant-bmc/README.md
RameshIyyar: > VPD collection will be occurring before redundancy is enabled.

I think VPD will collected during host IPL or after runtime.

During host IPL, FO is not allowed, so we’re fine. I believe VPD collection should happen after BSOC; otherwise, FO could be triggered during VPD collection, which might disrupt the flow. We can revisit this later if there's a valid use case to block FO during VPD collection.

RameshIyyar (approved):

redundant-bmc/README.md
spinler: Sounds good.  Ideally we'd just fold things like that into FailoversAllowed.

redundant-bmc/src/redundancy.cpp
spinler: Most of the checks here just prevent a failover from starting when it shouldn't have been called in the first place:
  failover in progress/imminent <---- FIP will be on D-Bus, maybe redfish? as you suggested maybe we could return a special error so they'd know they don't need to try again
    redundancyNotEnabled,  <--- shouldn't have been called
    fullSyncInProgress,           <--- probably full sync also in progress on active, so FOAllowed = false
    failoversNotAllowed,         <--- shouldn't have been called
    siblingDeadButRedundancyNotEnabled,  <------- shouldn't have been called because RedEnabled = false
    notAtReady,            <-- they could check for this and not call
    bmcNotPassive        < -- shouldn't have been called

I assume they'll get events when RedEnabled or FOAllowed changes.

PR #103, opened 2025-07-18
RBMC: Add service to wait for role target

spinler:
Add a new script wait-for-active-passive-target.sh that will wait for the role to be determined on D-Bus and then wait for either the obmc-bmc-active.target or obmc-bmc-passive.target to be started.

Also add a new service phosphor-wait-for-active-passive-target.service to call it.

This service will be linked to multi-user.target, so it can keep it from completing until the script is done.  This way, the BMC ready state can be held off until that is done.  This avoids the case where the BMC state says ready but it actually isn't ready because either the role is still unknown or all of the active services haven't been started yet.

Tested:

Waits for almost 2 minutes:
```
20:07:38 p11bmc systemd[1]: Starting Wait for obmc-bmc-active or obmc-bmc-passive target to be started
20:09:33 p11bmc wait-for-active-passive-target.sh[858]: BMC role determined, now waiting for obmc-bmc-passive.target to start
20:09:33 p11bmc wait-for-active-passive-target.sh[858]: Done waiting for obmc-bmc-passive.target
20:09:33 p11bmc systemd[1]: phosphor-wait-for-active-passive-target.service: Deactivated successfully.
20:09:33 p11bmc systemd[1]: Finished Wait for obmc-bmc-active or obmc-bmc-passive target to be started.
```

Change-Id: I4804b9a375b6267fca8a9f3f80a19bb44fbc313f

### DIFF
--- a/service_files/phosphor-wait-for-active-passive-target.service
+++ b/service_files/phosphor-wait-for-active-passive-target.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=Wait for obmc-bmc-active or obmc-bmc-passive target to be started
+After=xyz.openbmc_project.State.BMC.Redundancy.service
+
+[Service]
+ExecStart=/usr/libexec/phosphor-state-manager/wait-for-active-passive-target.sh
+Type=oneshot
+Restart=no
+
+[Install]
+RequiredBy=multi-user.target
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -21,6 +21,13 @@ anymore and so would default to active based on its position, and BMC 1 was
 active last time and would default to just restoring its role. Letting BMC 1 go
 first in this case would ensure BMC 0 remains passive.
 
+### Holding off BMC Ready
+
+A service file linked into multi-user.target that polls for the active and
+passive systemd targets to complete will prevent the system from reaching the
+BMC ready state before the role has been determined. On the active BMC this
+would also wait for all of the active applications to be started.
+
 ## Role Determination Rules
 
 The current rules for role determination are:
--- a/redundant-bmc/src/meson.build
+++ b/redundant-bmc/src/meson.build
@@ -43,3 +43,6 @@ executable(
     dependencies: [rbmc_deps, CLI11],
     install: true,
 )
+
+scripts = ['scripts/wait-for-active-passive-target.sh']
+install_data(scripts, install_mode: 'rwxr-xr-x', install_dir: execinstalldir)
--- a/redundant-bmc/src/scripts/wait-for-active-passive-target.sh
+++ b/redundant-bmc/src/scripts/wait-for-active-passive-target.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+# Wait for the Role property to get set to Active or Passive and
+# then wait for the corresponding obmc-bmc-<role>.target to start.
+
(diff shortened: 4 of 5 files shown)


RameshIyyar (approved):

redundant-bmc/src/wait-for-active-passive-target.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
geissonator: For org purposes, I'd probably move this script under a "scripts" subdir or something but we can always move later if we add more.

service_files/phosphor-wait-for-active-passive-target.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Wait for obmc-bmc-active or obmc-bmc-passive target to be started
+After=xyz.openbmc_project.State.BMC.Redundancy.service
+
+[Service]
+ExecStart=/usr/libexec/phosphor-state-manager/wait-for-active-passive-target.sh
+Type=oneshot
+
```
geissonator, line 9: "Restart=no" I don't think we'd want to go through the full timeout again with an auto restart on failure?
I was thinking we also may need TimeoutStartSec to set a larger timeout but looks like it's disabled by default with oneshot services.

service_files/phosphor-wait-for-active-passive-target.service
```diff
@@ -0,0 +1,10 @@
...
+[Install]
+WantedBy=multi-user.target
```
geissonator: So if this service fails we'll still go to Ready (only has WantedBy). Or will we also add this service to the critical service json and cause BMC Quiesced via that?

redundant-bmc/src/wait-for-active-passive-target.sh
spinler: I'll move it.

service_files/phosphor-wait-for-active-passive-target.service
spinler: Thanks, will add the Restart=no

service_files/phosphor-wait-for-active-passive-target.service
spinler: Hmm, I think quiescing is probably the right thing to do here, so that it doesn't just get permanently stuck in NotReady and people know to reboot if it's Quiesced.  I'll do some testing of that.

spinler: I tested that adding it to phosphor-service-monitor-default.json would result in the BMC being quiesced.  It worked with either the WantedBy or RequiredBy.  If I left RequiredBy but didn't add it to the JSON, then BMC state would just get stuck in NotReady.

I added a note into https://jsw.ibm.com/browse/PFEBMC-3525 to add that service when I add the other RBMC apps.  Going to try to hold off doing that until  I can get everything upstream.

redundant-bmc/src/wait-for-active-passive-target.sh
spinler: Done.

service_files/phosphor-wait-for-active-passive-target.service
spinler: Added Restart=no, and changed to RequiredBy

geissonator (approved):

PR #104, opened 2025-07-24
RBMC: Add toggleReset function to SiblingReset

spinler:
Put the calls to set the reset GPIO both to a 1 and a 0 into the same function, toggleReset, so the GPIO only has to be requested and released once.

The standalone assert and release reset functions are being kept for use in BMC card concurrent maintenance functionality.

Also add the SiblingReset class to the Providers class so code will have access to it.  It will be used during a failover to reset the original passive BMC so it can reboot and become passive.

Tested:
rbmctool --reset-sibling still works:

```
$ rbmctool --reset-sibling
<6> Toggling sibling reset GPIO

// journal:
rbmctool[1262]: Toggling sibling reset GPIO
phosphor-rbmc-state-manager[1143]: Sibling property Active changed
phosphor-rbmc-state-manager[1143]: Sibling BMC heartbeat lost
```

Change-Id: I0c15712c1720a06e6f72a08bdbfd8e3e937b4eff

### DIFF
--- a/redundant-bmc/src/rbmc_tool.cpp
+++ b/redundant-bmc/src/rbmc_tool.cpp
@@ -232,30 +232,18 @@ sdbusplus::async::task<> displayInfo(sdbusplus::async::context& ctx,
     std::cout << "\n";
 }
 
-void resetSiblingBMC()
+// NOLINTNEXTLINE
+sdbusplus::async::task<> resetSiblingBMC(sdbusplus::async::context& ctx)
 {
-    rbmc::SiblingResetImpl reset;
+    rbmc::SiblingResetImpl reset{ctx};
 
     try
     {
-        reset.assertReset();
+        co_await reset.toggleReset();
     }
     catch (const std::exception& e)
     {
         lg2::error("Failed asserting sibling reset: {ERROR}", "ERROR", e);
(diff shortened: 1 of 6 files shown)


redundant-bmc/src/rbmc_tool.cpp
```diff
@@ -394,7 +382,7 @@ int main(int argc, char** argv)
     }
     else if (resetSibling)
     {
-        resetSiblingBMC();
+        ctx.spawn(resetSiblingBMC(ctx));
```
RameshIyyar, line 385: I assume, we will wait to complete sibling reset.

geissonator (approved):
I'm good if Ramesh is

redundant-bmc/src/rbmc_tool.cpp
RameshIyyar: I noticed that `request_stop` is also triggered from `main`, and the context's `run` call will wait for all spawned tasks to complete, even after a stop is requested. This means `rbmctool` will finish any running tasks before stopping, and since `request_stop` prevents new tasks from being accepted and wait for running tasks.

RameshIyyar (approved):

PR #105, opened 2025-07-24
RBMC: Add FO-in-progress as no redundancy reason

spinler:
In a failover, the passive BMC will reset the original active BMC and then claim the active role itself and wait for the other BMC to come back as passive.

During that time, redundancy won't be enabled as the sibling BMC is in the middle of rebooting.  A new no redundancy reason of 'failover in progress' is being added that will be used during this time to make it obvious what is going on.

In addition, a new function 'disableRedundancyDueToFailover' is being added to RedundancyMgr that can be called to disable redundancy with this reason.  The reason would then persist on future calls to determine redundancy until the other new function, clearFailoverInProgress, is called.

Tested:
The new unit test that checks this reason passes.

Change-Id: Ia481ad1a74616625735c2a31f8efc737c58b1681

### DIFF
--- a/redundant-bmc/src/redundancy_mgr.cpp
+++ b/redundant-bmc/src/redundancy_mgr.cpp
@@ -115,7 +115,8 @@ redundancy::NoRedundancyReasons RedundancyMgr::getNoRedundancyReasons()
             services.getFWVersion() == sibling.getFWVersion().value_or(""),
         .manualDisable = manualDisable,
         .redundancyOffAtRuntimeStart = isRedundancyOffAtRuntime(),
-        .syncFailed = syncFailed};
+        .syncFailed = syncFailed,
+        .failoverInProgress = failoverInProgress};
 
     auto reasons = redundancy::getNoRedundancyReasons(input);
 
@@ -329,6 +330,12 @@ void RedundancyMgr::determineAndSetFailoversAllowed()
         // Already traced above.
         redundancyInterface.failovers_allowed(false);
     }
+}
+
+void RedundancyMgr::disableRedundancyDueToFailover()
+{
+    failoverInProgress = true;
+    determineAndSetRedundancy();
 }
 
 } // namespace rbmc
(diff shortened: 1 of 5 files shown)


redundant-bmc/src/redundancy_mgr.cpp
```diff
@@ -331,4 +332,10 @@ void RedundancyMgr::determineAndSetFailoversAllowed()
     }
 }
 
+void RedundancyMgr::disableRedundancyDueToFailover()
+{
+    failoverInProgress = true;
+    determineAndSetRedundancy();
```
RameshIyyar, line 338: I'm wondering if it's necessary to disable redundancy altogether. Wouldn't restricting FO be a more suitable approach?

spinler: Interesting idea. I will ponder that.

spinler: Do you think I should set FailoversAllowed to false on the active during the failover while the new passive is still rebooting, or just leave Redundancy=true and FailoversAllowed=true during all of  that, and then it would be evaluated again when passive comes back (or times out)?

RameshIyyar: > Do you think I should set FailoversAllowed to false on the active during the failover while the new passive is still rebooting

Yes, I would prefer that. However, there's a chance the new P-BMC might come back, and in that case, we could allow the FO again—similar to the AFO scenario. If it doesn't come back, then we'll need to disable redundancy. What I mean is, we should be able to reevaluate this situation during the FO process, right?

>just leave Redundancy=true and FailoversAllowed=true during all of that, and then it would be evaluated again when passive comes back (or times out)?

I don't think this will help in cases where there are double FO requests, right? We need to restrict FO requests during the FO process.

spinler: Will go direction of using FO Allowed, so will close this PR.

spinler:
Decided not to disable redundancy during the failover, so closing.

PR #106, opened 2025-07-29
RBMC: Add FO-in-progress check for FO not allowed

spinler:
In a failover, the passive BMC will reset the original active BMC and then claim the active role itself and wait for the other BMC to come back as passive.

During that time, the FailoversAllowed D-Bus property will be set to false to make it obvious to users that a failover isn't available then. When the passive BMC makes it back to ready state and the active BMC checks if redundancy can still be enabled, FailoversAllowed can be set back to true assuming nothing else is blocking it.

To handle this, this commit adds support for a 'failover-in-progress' reason for failovers to be disallowed.

It also adds a clearFailoversAllowedDuringFailover function on RedundancyMgr that will be called during the failover to call determineAndSetFailoversAllowed with failoverInProgress = true.

Later, RedundancyMgr::clearFailoverInProgress can be called to clear the failover in progress indication.

Tested:
New testcases pass.

Change-Id: I1107d1a67dba1c7ea003758d95055d2affcae4e7

### DIFF
--- a/redundant-bmc/src/redundancy.cpp
+++ b/redundant-bmc/src/redundancy.cpp
@@ -143,6 +143,13 @@ FailoversNotAllowedReasons getFailoversNotAllowedReasons(const Input& input)
         return reasons;
     }
 
+    if (input.failoverInProgress)
+    {
+        reasons.insert(failoverInProgress);
+        // No need to look for more reasons
+        return reasons;
+    }
+
     if (!input.fullSyncComplete)
     {
         reasons.insert(fullSyncNotComplete);
@@ -170,6 +177,9 @@ std::string getFailoversNotAllowedDescription(FailoversNotAllowedReason reason)
             break;
         case fullSyncNotComplete:
             desc = "A full sync hasn't been completed"s;
+            break;
+        case failoverInProgress:
+            desc = "A failover is in progress"s;
             break;
         case redundancyDisabled:
             desc = "Redundancy is disabled"s;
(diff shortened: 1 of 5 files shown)


redundant-bmc/src/redundancy_mgr.cpp
```diff
@@ -331,4 +332,10 @@ void RedundancyMgr::determineAndSetFailoversAllowed()
     }
 }
 
+void RedundancyMgr::calcFailoversAllowedDuringFailover()
```
geissonator: Why the "calc" in the function name and not "set". Seems like there's no calculations here, we're setting failover in progress and disabling fail overs?

spinler: The net result is failoversAllowed will get set to false on D-Bus.
Maybe I can go with 'clearFailoversAllowedDuringFailover'.

spinler: done.

redundant-bmc/src/redundancy.cpp
```diff
@@ -143,6 +143,13 @@ FailoversNotAllowedReasons getFailoversNotAllowedReasons(const Input& input)
         return reasons;
     }
 
+    if (input.failoverInProgress)
```
RameshIyyar, line 146: Commit message:

nit: typo `calcFailoversAllowedDuringFailover`

RameshIyyar (approved):

redundant-bmc/src/redundancy.cpp
spinler: thanks, fixed just that with the latest push.

PR #108, opened 2025-07-31
RBMC:  Implement the full failover

spinler:
Implement the fail over, broken up into a commit for the first part and a commit for the second.

### DIFF
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -208,7 +208,8 @@ Failovers aren't allowed when:
 2. The system is at some state other than off or runtime.
 3. `RedundancyEnabled` has changed to true but a full sync hasn't been
    completed.
-4. More coming.
+4. A failover is in progress.
+5. More coming.
 
 When failovers aren't allowed, rbmctool can be used to display the reasons why.
 
@@ -234,3 +235,21 @@ the request if any of the following are true.
    enabled. If it was last known to be enabled, a failover is allowed so that
    the remaining BMC can become active.
 1. The passive BMC is not in the `Ready` state.
+
+### Failover Sequence
+
+The failover sequence is:
+
+1. The `StartFailover` D-Bus method is called on the passive BMC daemon.
+1. The passive BMC checks to make sure
+   [it doesn't need to reject the request](#rejecting-a-failover-request). If it
+   does an error is returned via D-Bus.
+1. The passive BMC disables background syncs.
+1. The passive BMC sets its `FailoverImminent` property to true, and then waits
+   for 10 seconds to allow the other BMC to prepare for the failover assuming it
+   is alive.
+1. The passive BMC clears `FailoverImment` and sets `FailoverInProgress`.
+1. The passive BMC toggles the reset on the sibling BMC to reboot it.
+1. The passive BMC now switches its role to Active.
+1. The new active BMC sets `FailoversAllowed` to false.
+1. TODO: Remaining steps
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -252,4 +252,13 @@ The failover sequence is:
 1. The passive BMC toggles the reset on the sibling BMC to reboot it.
 1. The passive BMC now switches its role to Active.
 1. The new active BMC sets `FailoversAllowed` to false.
-1. TODO: Remaining steps
+1. The new active BMC takes over the hardware access bus. (TODO in code still)
+1. The new active BMC starts obmc-bmc-active.target. This starts all of the
+   active only services. It will not proceed until all active services have been
+   started which may take some time.
+1. The new active BMC waits for the sibling BMC to restart its heartbeat and get
+   to the BMC ready (or quiesced) state.
+1. The new active BMC clears `FailoverInProgress`.
+1. The new active BMC evaluates if redundancy can still be enabled, and if so it
+   issues a full sync. Otherwise it sets `RedundancyEnabled` to false.
+1. When the full sync is complete, `FailoversAllowed` is now set to true.
--- a/redundant-bmc/src/active_role_handler.cpp
+++ b/redundant-bmc/src/active_role_handler.cpp
@@ -165,4 +165,55 @@ auto ActiveRoleHandler::getFailoverBlockedReason(
     co_return fo_blocked::Reason::bmcNotPassive;
 }
 
+// NOLINTNEXTLINE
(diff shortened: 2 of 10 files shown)


redundant-bmc/README.md
geissonator, line 1: How come the "BMC Position:" was 0 in the test output? Shouldn't it have been the BMC in position 1?

redundant-bmc/src/manager.cpp
```diff
@@ -303,9 +303,65 @@ sdbusplus::async::task<> Manager::method_call(start_failover_t /* unused */,
...
+
+    // Stop handling as a passive BMC. This BMC no longer needs to
+    // watch for any changes from the one we're about to reset.
+    handler.reset();
+
+    redundancyInterface.failover_imminent(true);
+
+    // Wait to give the sibling a chance to react to seeing
+    // failover imminent before continuing.
+    co_await providers->getServices().doFailoverImminentDelay();
```
geissonator, line 334: I wonder. If this BMC (currently the passive, but in progress to be active) were to get rebooted right here, does our reboot flows handle that? Would the current active just stay the active?

geissonator (commented):
Nice, it's all coming together!

redundant-bmc/src/manager.cpp
spinler: Good question, what would happen during that whole flow is
1. passive sets failoverImminent
2. active (in future commit) sees that and stops background sync
3. passive BMC reboots before resetting active BMC
4. active BMC notices loss of passive BMC heartbeat 
5. passive BMC comes back as passive since there's already an active
6. active BMC sees passive BMC heartbeat come back and state = ready and does another full sync that restarts background sync when it checks redundancy again

A bigger window is if this BMC reboots after it already reset the original active BMC and becomes active, because then there are two rebooting BMCs that were previously active.  I have a future commit for that (the line 341 stuff).

redundant-bmc/README.md
spinler: I just must have already done 1 failover already to move BMC 0 from active to passive, and then for this output I did the failover on the passive BMC 0 that then moved to active.

geissonator (approved):

redundant-bmc/src/manager.cpp
```diff
@@ -303,9 +303,65 @@ sdbusplus::async::task<> Manager::method_call(start_failover_t /* unused */,
...
+    // NOLINTNEXTLINE(clang-analyzer-core.uninitialized.Branch)
+    co_await providers->getSyncInterface().disableBackgroundSync();
+
+    // Stop handling as a passive BMC. This BMC no longer needs to
+    // watch for any changes from the one we're about to reset.
+    handler.reset();
+
+    redundancyInterface.failover_imminent(true);
+
+    // Wait to give the sibling a chance to react to seeing
```
RameshIyyar, line 332: During the wait window, if the current Passive BMC reboots, it will return as Passive since it was already in that role and the Active BMC is still alive. However, if the reboot occurs after the wait window—or if the Active BMC itself is reset—the rebooted BMC will come back as Active. This is because the FO-InProgress property allows us to determine which BMC should assume the Active role when both are up.

The key difference is that in the latter case the FO process continues, while in the former case it does not. This raises a question: should the system (though this cannot happen within the BSOC context—HMC/PHYP would need to detect and request again) or the user be expected to re-trigger the FO?

Ideally, the system should recognise that the FO process was already in progress. Therefore, if the Passive BMC reboots during this time and comes back online, it should assume the Active role—since the FO was requested but interrupted due to the reboot. To support this, do we need to set the FO-InProgress property together with the FO-Imminent property?

redundant-bmc/src/active_role_handler.cpp
```diff
@@ -165,4 +165,55 @@ auto ActiveRoleHandler::getFailoverBlockedReason(
...
+    // Wait a bit to ensure the sibling reset is detected before we
+    // start waiting for it to come back. After the active target
+    // has real code in it that runs longer than a few seconds,
+    // this could be removed.
+    co_await providers.getSibling().pauseForHeartbeatChange();
+
+    // Wait for the sibling heartbeat to come back
+    co_await providers.getSibling().waitForSiblingUp();
+
+    // Wait for the sibling to get to Ready state so redundancy can
```
RameshIyyar: Just a suggestion: 

shall we mention like wait for sibling steady state to determine redundancy again?

spinler: I think that's what I have?  Or is the the 'again' that would sound better?  I can do that if so.

redundant-bmc/src/manager.cpp
spinler: I would just get rid of FailoverImminent if I were to move up FailoverInProgress, which I had been thinking about doing (though it would be a pain to remove).   But I'm not entirely sure how FailoverInProgress getting set earlier would help?  It may almost be better as it is now since if FailoverInProgress never gets set then they know they have a problem.  Or maybe it doesn't even matter and they can just keep an eye on the new active BMC and if it is lost then when it comes back they would check the role and handle it accordingly.

It is a pretty small window though, 10s so fairly unlikely to happen.

redundant-bmc/src/active_role_handler.cpp
RameshIyyar: I’m a bit confused about the difference between Ready and Steady states. I initially thought redundancy determination occurs once the system reaches the Ready state after seeing this statement, but I believe it actually happens when it reaches the Steady state which could be good or bad state, right?

redundant-bmc/src/manager.cpp
RameshIyyar: Yeah, it’s unlikely to happen.

Alternatively, we could rely on the FailoverImminent property to handle this case, so that it doesn’t require manual intervention.

Perhaps we should revisit this by simulating the scenario and then discuss how it should be addressed from an overall system perspective.

spinler: I opened a story to track discussing this with PHYP and the HMC.  From that we can see if we need to do anything more here.  Will resolv this comment.

redundant-bmc/src/active_role_handler.cpp
spinler: OK I see what you mean now.  I will update the comment to always say steady state, since if it hits Quiesced we will stop waiting.

spinler: Updated.

RameshIyyar (approved):

PR #110, opened 2025-08-26
RBMC: Preserve roles after failover reboot

spinler:
Consider the case where BMC 1 is far enough along in driving a failover that it has taken the active role, reset the other BMC, and is now waiting for it to come back as the passive BMC.

Normally when BMC 0 comes back it would just see the other BMC is already active and become passive.  However, if BMC 1 was rebooted sometime during its wait, then BMC 0 would come back first and wait for BMC 1 to catch up so they could determine their roles at roughly the same time.

At that point, there is nothing that directs BMC 0 to become passive as would be expected so it would just become active since it was previously active.

As for BMC 1, in the best case it would come up slightly behind BMC 0 enough to be able to claim its opposite role, but it may not so then it would end up also choosing the active role as it was previously active before the reboot.

To fix this so the roles are definitely preserved after the failover, do the following.
1. When FailoverInProgress is set to true on the new active BMC, also save it in the filesystem and then when its set back to false on D-Bus at the end of the failover remove it.
2. On the role determination that happens during startup, check if there is a persisted true value of FailoverInProgress.  If there is, that means this was a newly active BMC that was rebooted during the failover, so claim the active role.  After role determination is done, that persisted value can be removed.
3. Add support so the sibling BMC can obtain the FailoverInProgress value.
4. During role determination, if 'sibling failover in progress' is true, then claim the passive role as this BMC must be the newly passive one.

Tested:
Reboot the new active BMC in the middle of a failover, and get.

Passive BMC:
```
Local BMC
-----------------------------
Role:                 Passive <--------
BMC Position:         0
Redundancy Enabled:   true
BMC State:            Ready
Failovers Allowed:    true
Failover In Progress: false
FW Version Hash:      7D719923
Provisioned:          true
Role Reason:          Sibling was driving a failover <------
```

Active BMC:
```
Local BMC
-----------------------------
Role:                 Active <--------------------
BMC Position:         1
Redundancy Enabled:   true
BMC State:            NotReady
Failovers Allowed:    true
Failover In Progress: false
FW Version Hash:      7D719923
Provisioned:          true
Role Reason:          Newly active from failover <-------------
```

If it's is running behind, it can also show:

```
Local BMC
-----------------------------
Role:                 Active
BMC Position:         1
Redundancy Enabled:   true
BMC State:            Ready
Failovers Allowed:    true
Failover In Progress: false
FW Version Hash:      7D719923
Provisioned:          true
Role Reason:          Sibling is already passive
```

Change-Id: I634932451fb737fde1e2efedb44bbcb27a148a1c

### DIFF
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -37,6 +37,9 @@ The current rules for role determination are:
 1. If the sibling isn't provisioned, choose active.
 1. If the sibling is already passive, choose active.
 1. If the sibling is already active, choose passive.
+1. If there was previously a failover in progress, choose active. See
+   [below for more details](#reboots-in-the-middle-of-a-failover).
+1. If the sibling has a failover in progress, choose passive.
 1. If the previous role isn't unknown, choose that assuming it wasn't passive
    just due to an error.
 1. Finally, if this BMC's position is zero choose active, otherwise passive.
@@ -269,3 +272,22 @@ The failover sequence is:
 1. The new active BMC evaluates if redundancy can still be enabled, and if so it
    issues a full sync. Otherwise it sets `RedundancyEnabled` to false.
 1. When the full sync is complete, `FailoversAllowed` is now set to true.
+
+### Reboots in the middle of a failover
+
+If the new active BMC is rebooted in the middle of a failover while the other
+BMC is somewhere in its reboot, the roles should be preserved with the new
+active BMC staying active.
+
+Preserving the roles is accomplished by:
+
+- On the new active BMC at the start of the failover save the failover in
+  progress indication in the filesystem. On every startup, look for that
+  indication and use it in the
+  [role determination calculation](#role-determination-rules). It will be
+  cleared from the filesystem after a successful failover or after it was used
+  in role determination.
+
+- On the new passive BMC during role determination obtain the failover in
+  progress value from the other BMC. If it is true, then choose the passive
+  role.
--- a/redundant-bmc/src/manager.cpp
+++ b/redundant-bmc/src/manager.cpp
@@ -48,6 +48,24 @@ Manager::Manager(sdbusplus::async::context& ctx,
                    "ERROR", e);
     }
 
+    try
+    {
+        // Restore FailoverInProgress on D-Bus
+        auto failoverInProgress =
+            data::read<bool>(data::key::failoverInProgress).value_or(false);
+        if (failoverInProgress)
+        {
+            lg2::info("Failover was previously in progress");
+        }
+
+        redundancyInterface.failover_in_progress(failoverInProgress);
+    }
+    catch (const std::exception& e)
+    {
+        lg2::error("Failed trying to obtain failover-in-progress: {ERROR}",
+                   "ERROR", e);
+    }
+
     // emit the Failover interfaces added signal
     emit_added();
 
@@ -89,9 +107,38 @@ sdbusplus::async::task<> Manager::startup()
         updateRole(determineRole());
     }
 
-    spawnRoleHandler();
+    co_await postStartupClearFOInProgress();
 
-    co_return;
+    spawnRoleHandler();
+}
+
+// NOLINTNEXTLINE
+sdbusplus::async::task<> Manager::postStartupClearFOInProgress()
+{
+    if (!redundancyInterface.failover_in_progress())
+    {
+        co_return;
+    }
+
+    // If true, this property is used by both BMCs to determine their roles.
+    // This BMC will have already used it.  Check the other BMC already has
+    // as well and then it can be set to false and removed from the persistent
+    // data.
+
+    lg2::info(
+        "Waiting for sibling to get role and then clearing failover in progress");
+
+    co_await providers->getSibling().waitForSiblingRole();
+
+    redundancyInterface.failover_in_progress(false);
+    try
+    {
+        data::remove(data::key::failoverInProgress);
+    }
+    catch (const std::exception& e)
+    {
+        lg2::error("Failed removing failover-in-progress: {ERROR}", "ERROR", e);
+    }
 }
 
 void Manager::spawnRoleHandler()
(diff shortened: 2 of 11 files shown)


redundant-bmc/src/sibling_impl.hpp
```diff
@@ -168,6 +168,21 @@ class SiblingImpl : public Sibling
...
+    /**
+     * @brief Returns if the sibling has a failover in progress
+     *
+     * @return - If in progress, or nullopt if not available
+     */
+    std::optional<bool> getFailoverInProgress() const override
+    {
+        if (getInterfacePresent() && hasHeartbeat())
+        {
+            return redundancy.failoverInProgress;
```
geissonator, line 180: I'm missing something here, how do we actually get the info from the sibling on whether it had the fail over in progress file? Is it something in the op-proc-control CFAM interface?

redundant-bmc/README.md
```diff
@@ -269,3 +272,22 @@ The failover sequence is:
 1. The new active BMC evaluates if redundancy can still be enabled, and if so it
    issues a full sync. Otherwise it sets `RedundancyEnabled` to false.
 1. When the full sync is complete, `FailoversAllowed` is now set to true.
+
+### Reboots in the middle of a failover
+
+If the new active BMC is rebooted in the middle of a failover while the other
```
geissonator, line 278: I wonder what is expected if something about the new Active BMC going to Active is causing the BMC reboot (i.e. some bug in the BMC to power supplies or BMC to CEC chips causing a kernel panic). Would it be possible for the HMC to log into the new Passive and force another fail over back? I'm just a little worried about an infinite reboot scenario (but I don't want to write any complicated logic to try and handle it).

spinler: Yea, I think they should be able to do that, though if they failed over for a reason the original active may not be doing so hot either.

redundant-bmc/src/sibling_impl.hpp
spinler: Yea, this CFAM daemon code here putting that field on D-Bus: https://github.com/ibm-openbmc/openpower-proc-control/blob/1120/rbmc-cfam-daemon/src/sibling_bmc.cpp#L75

And this code from right above is getting it from D-Bus:
```
    it = propertyMap.find("FailoverInProgress");
    if (it != propertyMap.end())
    {
        redundancy.failoverInProgress = std::get<bool>(it->second);
    }
```

geissonator (approved):

redundant-bmc/README.md
```diff
@@ -37,6 +37,9 @@ The current rules for role determination are:
 1. If the sibling isn't provisioned, choose active.
```
RameshIyyar, line 37: In the commit message FO-Inprogress and FO-Allowed value are correct? or its just mock up?

redundant-bmc/README.md
```diff
@@ -269,3 +272,22 @@ The failover sequence is:
...
+If the new active BMC is rebooted in the middle of a failover while the other
+BMC is somewhere in its reboot, the roles should be preserved with the new
+active BMC staying active.
+
+Preserving the roles is accomplished by:
+
+- On the new active BMC at the start of the failover save the failover in
+  progress indication in the filesystem. On every startup, look for that
+  indication and use it in the
+  [role determination calculation](#role-determination-rules). It ill be cleared
```
RameshIyyar: typo: It ill be -> It will be

redundant-bmc/README.md
RameshIyyar: I think this could lead to an infinite reboot loop (including retries), right? Unless the sibling BMC can’t be used to boot the host so triggered a FO—in that case we’d need a way to disable redundancy, correct? like some of kind of hard failure in BSOC?

redundant-bmc/src/manager.cpp
```diff
@@ -89,9 +107,38 @@ sdbusplus::async::task<> Manager::startup()
...
+
+    lg2::info(
+        "Waiting for sibling to get role and then clearing failover in progress");
+
+    co_await providers->getSibling().waitForSiblingRole();
+
+    redundancyInterface.failover_in_progress(false);
+    try
+    {
+        data::remove(data::key::failoverInProgress);
```
RameshIyyar, line 136: I’m wondering if there could be issues when syncing RBMC persisted data to another BMC. Since we don’t sync during FO, if one or both BMCs reboot, we would still rely on the unsynced persisted FO-Inprogress value for role claiming. I just want to confirm if this flow, or do you think we should separate this data into a file that isn’t synced?

redundant-bmc/src/manager.cpp
```diff
@@ -338,14 +389,23 @@ sdbusplus::async::task<> Manager::doFailoverFromPassive()
...
-    // so it can be used to know that this BMC should be active
-    // if rebooted in the middle of this.
-
     // Reset the active so it can come back as passive.
     // If this were to throw, let it restart the app.
     co_await providers->getSiblingReset().toggleReset();
 
+    try
+    {
+        // Now that its past the reset, save the failover in progress
```
RameshIyyar, line 398: I think the reset should be quick, but do you think we should persist immediately after the FO-Inprogress property is updated?

redundant-bmc/src/manager.cpp
spinler: This app's persistent data file data.json isn't synced, it's just the local value.

redundant-bmc/src/manager.cpp
spinler: Functionally it wouldn't matter either way.  Would you prefer it be done there?

redundant-bmc/README.md
spinler: done.

redundant-bmc/README.md
spinler: Do you mean here?
```
    Local BMC
    -----------------------------
    Role:                 Passive 
    BMC Position:         0
    Redundancy Enabled:   true
    BMC State:            Ready
    Failovers Allowed:    true
    Failover In Progress: false
    FW Version Hash:      7D719923
    Provisioned:          true
    Role Reason:          Sibling was driving a failover 
```
That was taken at steady state, after everything was back again.

redundant-bmc/src/manager.cpp
RameshIyyar: not really :)

RameshIyyar (approved):

redundant-bmc/README.md
spinler: This scenario is being discussed further in slack.

PR #111, opened 2025-09-04
Flush journal and stop background sync before failover reset

spinler:
On the existing active BMC, watch for the failover imminent indication and stop background sync and flush the journal to disk.

### DIFF
--- a/redundant-bmc/src/services_impl.cpp
+++ b/redundant-bmc/src/services_impl.cpp
@@ -65,6 +65,88 @@ sdbusplus::async::task<std::string> getService(sdbusplus::async::context& ctx,
     co_return object.begin()->first;
 }
 
+/**
+ * @brief Run a shell command asynchronously
+ *
+ * Use the pipe()/fork() paradigm to fork off a child process to run
+ * the command. The child will write the command RC to its FD obtained
+ * from pipe() and exit. Meanwhile the parent will use async::fdio to
+ * asynchronously wait for that rc to show up in its read FD it had
+ * obtained from pipe().
+ */
+// NOLINTNEXTLINE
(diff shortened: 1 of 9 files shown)


redundant-bmc/src/services_impl.cpp
```diff
@@ -65,6 +65,86 @@ sdbusplus::async::task<std::string> getService(sdbusplus::async::context& ctx,
...
+        // Child
+
+        // Close the read pipe
+        close(pipeFDs[0]);
+
+        // NOLINTNEXTLINE(cert-env33-c)
+        int rc = std::system(cmd.c_str());
+
+        // Write the return code to the write pipe
+        ssize_t s = write(pipeFDs[1], &rc, sizeof(rc));
```
RameshIyyar: If the intent is to capture the actual exit status of `journalctl --sync` for debugging, relying on the return value from `std::system` is insufficient. The correct approach is to extract the exit code via `WEXITSTATUS`, for example:

```
int rc = std::system(cmd.c_str());
int exitCode = (rc == -1) ? -1 :
               (WIFEXITED(rc) ? WEXITSTATUS(rc) : -1);
write(pipeFDs[1], &exitCode, sizeof(exitCode));
```

Additionally, this results in two forks: one explicitly at line 92 and another internally from `std::system`. Given the available 10-second window, the extra overhead should not pose a performance concern.

spinler: Good suggestion, will do that.

spinler: Done.  Only change in the new push.

RameshIyyar (approved):

PR #113, opened 2025-09-19
RBMC: Get BMC position from inventory

spinler:
Each BMC needs to know its own position, and was previously just getting it from a uboot environment variable set by the simulator.

Now get the position from the real place, which is the Position property on the xyz.openbmc_project.Inventory.Decorator.Position interface on /xyz/openbmc_project/inventory/system.

If the value is the 'maxint' value it means the position could not be obtained due to a hardware issue.  In that case, or if there is some D-bus failure getting the position, the BMC will have to be passive because a known position is required to become active.

Tested:
Position is obtained:
```
$ rbmctool -de |grep Position
BMC Position:         0

$ rbmctool -de | grep Position
BMC Position:         1
```

Change-Id: If2855cec9dbaaca0a233915ae1ad5490c15bb7c8

### DIFF
--- a/redundant-bmc/src/services_impl.hpp
+++ b/redundant-bmc/src/services_impl.hpp
@@ -41,9 +41,10 @@ class ServicesImpl : public Services
     /**
      * @brief Returns this BMC's position.
      *
-     * @return - The position
+     * @return - The position if it can be obtained
      */
-    size_t getBMCPosition() const override;
+    sdbusplus::async::task<std::optional<size_t>> getBMCPosition()
+        const override;
 
     /**
      * @brief Starts a systemd unit
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -53,6 +53,7 @@ There are some error cases that require that the BMC is passive regardless of
 what the sibling is doing. These are:
 
 1. The BMC is not provisioned.
+1. The BMC position cannot be determined.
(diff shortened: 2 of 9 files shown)


redundant-bmc/src/services_impl.hpp
```diff
@@ -41,9 +41,12 @@ class ServicesImpl : public Services
     /**
      * @brief Returns this BMC's position.
      *
-     * @return - The position
+     * @param[in] silent - If error traces should be printed
```
geissonator: Is this "silent" concept new? Why wouldn't we want to print out error messages in both cases?

spinler: I noticed that if the position is bad when rbmctool -de prints it, it interlaces the lg2::error mesages right in the middle of the rbmctool output since lg2 writes to the console and thought it looked ugly.  I don't care too much though, and can remove the silent stuff if you prefer.

geissonator: Yeah, I'm really not a fan of passing around booleans and doing if/elses on them (maybe it's from my fsp-trace days). If anything we could do a marcro wrapper around them that hides it, but it's error path and a debug tool so I don't see anything wrong with just printing it out.

spinler: done.

geissonator (approved):

RameshIyyar (approved):

PR #114, opened 2025-10-07
RBMC: Use Available prop to get sib BMC presence

spinler:
Previously, the Sibling::isBMCPresent check was just hardcoded to return true, which of course doesn't work if the BMC may not be present.

Ideally there would be GPIOs that could be used for this, however not all of IBM's redundant BMC systems have them, and the one that will isn't far enough along to code against.

So for now, just look at the Availability interface on the state/bmc1 object path which reflects if the communication path to the sibling BMC was detected.

When more information is known about the presence GPIO, it can be checked first and then code can fall back to the availability check if the GPIO read fails due to a hardware error or if the system doesn't have a GPIO.

Add this new interface and property to the Sibling handling of the other interfaces on that object path which already watch for all the D-Bus signals to track changes, and return the value of Available in the isBMCPresent() function.

The interface will be present as soon as the sibling daemon starts up regardless of if the BMC is responding or not.

Tested:

When available, works the same as before.

When not available:
```
Local BMC
-----------------------------
Role:                 Active
BMC Position:         0
Redundancy Enabled:   false
BMC State:            Ready
Failovers Allowed:    false
Failover In Progress: false
FW Version Hash:      76A7D135
Provisioned:          true
Role Reason:          Sibling not alive
    Sibling is missing

Sibling BMC
-----------------------------
Sibling data not available
```

Change-Id: I5bdfa18d8e15b261c75e9f55a605d58cc115617c

### DIFF
--- a/redundant-bmc/src/sibling_impl.hpp
+++ b/redundant-bmc/src/sibling_impl.hpp
@@ -44,7 +44,8 @@ class SiblingImpl : public Sibling
      */
     bool alive() const override
     {
-        return version.present && redundancy.present && bmcState.present;
+        return version.present && redundancy.present && bmcState.present &&
+               availability.present;
     }
 
     /**
@@ -198,7 +199,10 @@ class SiblingImpl : public Sibling
      *
      * @return bool - if present
      */
-    bool isBMCPresent() override;
+    bool isBMCPresent() override
+    {
+        return availability.available;
+    }
 
     /**
      * @brief Waits for up to 10 minutes for the sibling BMC to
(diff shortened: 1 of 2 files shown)


spinler:
CI Fail was:
```
docker: error during connect: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.47/containers/create": EOF.
See 'docker run --help'.
```
will try again.

spinler:
jenkins run tests please

RameshIyyar (approved):

redundant-bmc/src/sibling_impl.hpp
```diff
@@ -44,7 +44,8 @@ class SiblingImpl : public Sibling
      */
     bool alive() const override
     {
-        return version.present && redundancy.present && bmcState.present;
+        return version.present && redundancy.present && bmcState.present &&
+               availability.present;
```
geissonator, line 48: It seems a bit counter intuitive that .present is checked in the alive() interface and .available is checked in the isBMCPresent() call. What's the difference between these two? Kind of confusing.

spinler: The CFAM daemon will
1. Set Available based on if the /dev device for the sibling CFAM is there.  This is the analogue for present for now.
2. Create/remove the other sibling D-Bus interfaces based on if the sibling BMC's heartbeat is good.  This is 'alive'.

So here, isBMCPresent() checks available (for huygens we should be able to check the GPIOs first)
and alive() checks for the existence of the interfaces that are only there if the sibling BMC is alive (good heartbeat).

I probably didn't need to add the availability.present check in alive(), but I don't think it should hurt anything.

PR #117, opened 2025-10-16
RBMC: Log failover requesters/timestamps

spinler:
The 'requester' parameter of the StartFailover method says who is requesting the failover.  Save that and the timestamp the failover was requested in a file so that it can be displayed by rbmctool and also captured in BMC dumps.

A new 'logFailover' function is used for this. It is called by the BMC driving the failover after the point of no return which is after the original active BMC is reset.  It writes to a JSON file that keeps the most recent 10 failovers.

The file is in the same directory the existing 'data.json' file is in, and is named bmcN_failovers, where N is the BMC position.  To get a full view of every failover in the system from a single BMC, the files would need to be synced so each BMC has the file for both positions.

The 'rbmctool -de' option will now display the requester and timestamp of the most recent failover when run on the active BMC. It doesn't do this on the passive BMC because the most recent failover saved on that BMC isn't the actual most recent one, as that would be on the current active BMC.  If the files are synced in the future, then possibly it could use the file synced over from the active BMC.

Tested:

Most recent failover is shown:
```
Local BMC
-----------------------------
Role:                 Active
BMC Position:         1
Redundancy Enabled:   true
BMC State:            Ready
Failovers Allowed:    true
Failover In Progress: false
FW Version Hash:      7D719923
Provisioned:          true
Role Reason:          Failover
Last failover driven by this BMC:
    Requester: Tool                    <----
    Timestamp: 2025-10-08 17:14:24 UTC <----

```

And the file itself:
```
$ cat /var/lib/phosphor-state-manager/redundant-bmc/bmc1_failovers
[
    {
        "requester": "Tool",
        "timestamp": "2025-10-08 16:23:54 UTC"
    },
    {
        "requester": "Tool",
        "timestamp": "2025-10-08 17:14:24 UTC"
    }
]
```

Also, new testcases pass.

Change-Id: I1fc538c0f5fcb3e14f7b1159d23f19717c350820

### DIFF
--- a/redundant-bmc/src/persistent_data.cpp
+++ b/redundant-bmc/src/persistent_data.cpp
@@ -9,6 +9,9 @@
 
 namespace data
 {
+
+constexpr size_t maxFailoverLogEntries = 10;
+
 namespace util
 {
 
@@ -47,6 +50,25 @@ void writeFile(const nlohmann::json& json, const std::filesystem::path& path)
     }
 }
 
+static std::filesystem::path makeFailoverLogFilePath(
+    const std::filesystem::path& dirPath, size_t bmcPos)
+{
+    std::string filename = "bmc" + std::to_string(bmcPos) + "_failovers";
+    return dirPath / filename;
+}
+
+static std::string makeTimestampString(const time_t& timestamp)
+{
+    tm utcTime{};
+    gmtime_r(&timestamp, &utcTime);
+
+    std::string timeBuf(50, '\0');
+    auto size = strftime(timeBuf.data(), timeBuf.size(),
+                         "%Y-%m-%d %H:%M:%S UTC", &utcTime);
+    timeBuf.resize(size);
+    return timeBuf;
+}
+
 } // namespace util
 
 void remove(std::string_view name, const std::filesystem::path& path)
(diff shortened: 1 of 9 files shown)


spinler:
This will have to go in at the same time as the PDI change to add the new Requester arg: https://github.com/ibm-openbmc/phosphor-dbus-interfaces/pull/192.  Until then, it won't pass CI.

```
../redundant-bmc/src/persistent_data.hpp:19:65: error: ‘Requester’ in ‘struct sdbusplus::common::xyz::openbmc_project::control::Failover’ does not name a type
   19 |     sdbusplus::common::xyz::openbmc_project::control::Failover::Requester;
      |                                                                 ^~~~~~~~~
../redundant-bmc/src/persistent_data.hpp:143:18: error: ‘Requester’ has not been declared
  143 |                  Requester requester, const time_t& timestamp);
```

RameshIyyar (approved):
Nice.

redundant-bmc/src/persistent_data.cpp
```diff
@@ -63,4 +85,68 @@ void remove(std::string_view name, const std::filesystem::path& path)
...
+        // If at the max number of entries, drop the oldest.
+        if (logs->size() == maxFailoverLogEntries)
+        {
+            logs->erase(logs->begin());
+        }
+
+        // Just save the last segment of the enum string, e.g. 'Host'.
+        auto r = sdbusplus::common::xyz::openbmc_project::control::Failover::
+            convertRequesterToString(requester);
+        r = r.substr(r.find_last_of('.') + 1);
```
geissonator, line 116: ok, string stuff always makes my head hurt. Why would this not be a "- 1" to go back one from the "."?

spinler: Say you have "1.2".
find_last_of('.') will return index 1
substr starting at index 1 would then give ".2"
Rather we want 1 past the position of the '.'

geissonator (approved):

PR #120, opened 2025-11-13
Go to passive is system VPD collection isn't successful.

spinler:
A redundant BMC system that utilizes phosphor-inventory-manager has a
system inventory progress that must have a value of 'Completed' in order
for the BMC to be active. It uses the 'Common.Progress' interface and is
on the same object path as the 'Item.System' inventory interface.

To handle this, do the following:
1. Wait up to 3 minutes for the 'Item.System' interface to show up on
   D-Bus, and save that object path.
2. Start using that object path in the code that reads the BMC position,
   as that is at the same place.
3. Check for that 'Common.Progress' interface.  If it isn't present,
   assume the system didn't implement it and consider it a success.
4. Wait up to 3 minutes for the Status property on that interface to
   reach either Completed, Failed, or Aborted.
5. If it doesn't reach that state in 3 minutes or if it is Failed or
   Aborted, consider it a failure.
6. On failures, force the BMC to become passive.


redundant-bmc/src/role_determination.cpp
```diff
@@ -115,6 +115,9 @@ std::string getRoleReasonDescription(RoleReason reason)
         case RoleReason::unknownBMCPosition:
             desc = "Cannot determine BMC position"s;
             break;
+        case RoleReason::systemInventoryFailed:
+            desc = "System inventory failed";
```
RameshIyyar: System inventory is not available?

redundant-bmc/src/role_determination.hpp
```diff
@@ -45,7 +45,8 @@ enum class RoleReason
     failover,
     failoverInProgress,
     siblingFailoverInProgress,
-    unknownBMCPosition
+    unknownBMCPosition,
+    systemInventoryFailed
```
RameshIyyar: systemInventoryNotAvailable?

redundant-bmc/src/services_impl.cpp
```diff
@@ -69,6 +73,34 @@ sdbusplus::async::task<std::string> getService(sdbusplus::async::context& ctx,
...
+    sdbusplus::async::context& ctx)
+{
+    auto mapper = ObjectMapper(ctx)
+                      .service(ObjectMapper::default_service)
+                      .path(ObjectMapper::instance_path);
+
+    std::vector<std::string> systemIface{SystemInv::interface};
+
+    auto objects = co_await mapper.get_sub_tree(
+        "/xyz/openbmc_project/inventory", 0, systemIface);
```
RameshIyyar, line 86: Good to add Position interface?

redundant-bmc/src/services_impl.cpp
```diff
@@ -69,6 +73,34 @@ sdbusplus::async::task<std::string> getService(sdbusplus::async::context& ctx,
...
+
+    if (objects.empty())
+    {
+        throw std::runtime_error("No system inventory object found");
+    }
+
+    // Until there is a reason to expect more, check
+    // that there is just one System interface.
+    if (objects.size() != 1)
```
RameshIyyar, line 95: wont we get two objects, one from the PIM and another one from EM?

redundant-bmc/src/services_impl.cpp
```diff
@@ -573,4 +620,99 @@ sdbusplus::async::task<> ServicesImpl::flushJournal() const
...
+sdbusplus::async::task<bool> ServicesImpl::checkSystemInventoryStatus()
+{
+    if (systemInvPath.empty())
+    {
+        lg2::error("There is no system inventory object");
+        co_return false;
+    }
+
+    using namespace std::chrono_literals;
+    auto end = std::chrono::steady_clock::now() + 3min;
```
RameshIyyar, line 671: Do you think matcher, will save us from this timeout? and I guess we need to timeout for matcher too, we cannot wait for PC...

spinler: Right, I think we'd still need to have a timeout on the matches. And I think I would need to handle 2 signals, properties changed and interfaces added, so I think not using them is a bit easier.

redundant-bmc/src/services_impl.cpp
spinler: To get a System interface from EM, someone would have to explicitly put it in a config file.  I can't think of a reason someone would want one in both PIM and EM, but if they did, hopefully then they'd also have a way to figure out which one to use here.

redundant-bmc/src/services_impl.cpp
spinler: Down below it's already looking up the service that has the Position interface, so I don't think it is necessary here.  Let me know if I'm missing something though.

redundant-bmc/src/role_determination.cpp
spinler: Sure. Done.

redundant-bmc/src/role_determination.hpp
spinler: Done.

spinler:
Jenkins run tests please

redundant-bmc/src/services_impl.cpp
RameshIyyar: What I was thinking is that we could avoid one D-Bus call (LN422) if we also pass the Position interface here.

redundant-bmc/src/services_impl.cpp
RameshIyyar: Hmm, recently I was trying to get the system inventory service and noticed that it’s EM. I also tried using the Position interface. Let me double-check this again.

geissonator (commented):
"The IBM system that has redundant BMCs has a system inventory progress
that must have a value of 'Completed' in order for the BMC to be active."

This in one of the commit messages is going to be a bit rough to upstream  (specific mention of an IBM system). Can we make it a bit more generic?

redundant-bmc/src/manager.cpp
```diff
@@ -229,17 +229,24 @@ sdbusplus::async::task<std::optional<role_determination::RoleInfo>>
 {
     using namespace role_determination;
 
+    // An unprovisioned BMC cannot be active.
+    if (!providers->getServices().getProvisioned())
```
geissonator, line 233: This seems unrelated to the commit?

spinler: I just re-ordered the checks here to make them match the documentation in the README (lines 55-57), so I'd prefer to keep in this commit.

spinler:
> "The IBM system that has redundant BMCs has a system inventory progress that must have a value of 'Completed' in order for the BMC to be active."
> 
> This in one of the commit messages is going to be a bit rough to upstream (specific mention of an IBM system). Can we make it a bit more generic?

Updated the commit message, and also the PR description to make it more generic.

geissonator (approved):

redundant-bmc/src/services_impl.cpp
```diff
@@ -573,4 +620,99 @@ sdbusplus::async::task<> ServicesImpl::flushJournal() const
...
+
+    while (std::chrono::steady_clock::now() < end)
+    {
+        InvProgress::OperationStatus status;
+
+        try
+        {
+            auto service = co_await util::getService(ctx, systemInvPath,
```
RameshIyyar: Do we need to this for every iteration?

redundant-bmc/src/services_impl.cpp
```diff
@@ -573,4 +620,99 @@ sdbusplus::async::task<> ServicesImpl::flushJournal() const
...
+        else if ((status == InvProgress::OperationStatus::Failed) ||
+                 (status == InvProgress::OperationStatus::Aborted))
+        {
+            lg2::error("System inventory failed with status {STATUS}", "STATUS",
+                       status);
+            co_return false;
+        }
+        else
+        {
+            lg2::info("System inventory status isn't complete yet: {STATUS}",
```
RameshIyyar: > System inventory status isn't complete yet: xyz.openbmc_project.Common.Progress.OperationStatus.NotStarted
System inventory status isn't complete yet: xyz.openbmc_project.Common.Progress.OperationStatus.NotStarted
System inventory status isn't complete yet: xyz.openbmc_project.Common.Progress.OperationStatus.NotStarted


Do we need this trace for every 1s?

redundant-bmc/src/services_impl.cpp
spinler: Will fix.

redundant-bmc/src/services_impl.cpp
spinler: I will make it smarter.

RameshIyyar (approved):

PR #121, opened 2025-12-01
Reapply "phosphor-discover-system-state: set after multi-user.target"

geissonator:
This reverts commit 0a7f8f9a1a1a9f856a9777aed12577ceab92f1a1.

It turns out that if you have the only-allow-boot-when-bmc-ready feature enabled (which we do on all IBM systems), then it requires we have the APR service configured to run after multi-user.target.

Otherwise, the APR service must complete before we can get to BMC Ready and even with its 30s delay we end up not being at BMC Ready when the power on is requested, which means it immediately fails.

Will need to revisit this problem in our 1210 release to support redundant BMC machines.


spinler (approved):
This would break skiboards since it'd no longer be able to get to ready, but nobody should be using it anymore there.

PR #123, opened 2025-12-09
remove installs on templated services and targets

geissonator:
New yocto is getting more stringent on installation of templated service and targets.

Templated services are installed by bitbake recipes so there's no need to have the WantedBy/RequiredBy in those services.

Here's a summary of the error we get for packages which have templated services doing an install:
```
ERROR: obmc-phosphor-image-1.0-r0 do_rootfs: Postinstall scriptlets of
['<package>' 'inst returned 1, marking as unpacked only, configuration
required on target']
If the intention is to defer them to first boot,
then please place them into pkg_postinst_ontarget:${PN} ().
Deferring to first boot via 'exit 1' is no longer supported.
```

Tested:
- Confirmed image builds with new yocto and services are still installed as expected

Change-Id: I56d9ac7a73dfd6bdaa05eba39d890a8af3605d18

### DIFF
--- a/host_condition_gpio/phosphor-host-condition-gpio@.service
+++ b/host_condition_gpio/phosphor-host-condition-gpio@.service
@@ -8,4 +8,4 @@ Type=dbus
 BusName=xyz.openbmc_project.State.HostCondition.Gpio%i
 
 [Install]
-WantedBy=multi-user.target
+# Handled by bitbake recipe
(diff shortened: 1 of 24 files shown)


geissonator:
Upstream: https://gerrit.openbmc.org/c/openbmc/phosphor-state-manager/+/85742

PR #124, opened 2025-12-11
RBMC: Check if the passive BMC has a reason for no redundancy

spinler:
In some cases, the passive BMC cannot ever be active. For example, if it couldn't read system VPD.  Instead of having a new bit in the CFAM for every possible reason that that active has to check for, this commit makes use of a new summary bit that says the passive BMC can never be active.  

Include that check when enabling redundancy.


spinler:
The PR https://github.ibm.com/openbmc/openpower-proc-control/pull/81 is a prereq to put the bit in the CFAM.

RameshIyyar (approved):

spinler:
Will refactor this around 'not redundancy capable' to match what will be done with https://github.ibm.com/openbmc/openpower-proc-control/pull/81.

PR #125, opened 2026-01-16
RBMC: Check D-Bus service name for sib service

spinler:
For a BMC to be active, the sibling BMC application must be up and running so that it can get info about the sibling BMC.

Simplify the check to see if the sibling service is running by checking if it is providing the State.Decorator.Availability interface on the '/xyz/openbmc_project/state/bmc1' path, which will be present by the time it requests its bus name.  Only after that is done would systemd start the RBMC daemon via the service file dependency.

The previous method didn't always work in some of the error cases when the sibling service was in the middle of crashing and  restarting, which could lead to two active BMCs.

Tested:

When the sibling service is dead the Passive role is properly chosen:

```
Local BMC
-----------------------------
Role:                 Passive
BMC Position:         0
Redundancy Enabled:   false
BMC State:            Ready
Failovers Allowed:    false
Failover In Progress: false
FW Version Hash:      C5EABAEA
Provisioned:          true
Role Reason:          Sibling BMC service is not running
Cannot Be Active:     true

Sibling BMC
-----------------------------
Sibling BMC data not available
```

Change-Id: I193e42c8a5122827c8060edddd787536b307d534

### DIFF
--- a/redundant-bmc/src/manager.cpp
+++ b/redundant-bmc/src/manager.cpp
@@ -249,17 +249,11 @@ sdbusplus::async::task<std::optional<role_determination::RoleInfo>>
                            RoleReason::systemInventoryNotAvailable};
     }
 
-    // The sibling service must be up and running.
-    if (!providers->getSibling().alive())
+    // If the sibling service isn't on D-Bus, the BMC can't be active.
+    if (providers->getSibling().getServiceName().empty())
     {
-        auto state =
-            co_await providers->getServices().getUnitState(Sibling::unitName);
-        if ((state != "active") && (state != "activating"))
-        {
-            lg2::info("Sibling service state is {STATE}", "STATE", state);
-            co_return RoleInfo{Role::Passive,
-                               RoleReason::siblingServiceNotRunning};
-        }
+        lg2::error("Sibling service is not running");
+        co_return RoleInfo{Role::Passive, RoleReason::siblingServiceNotRunning};
     }
 
     co_return std::nullopt;
--- a/redundant-bmc/src/sibling_impl.cpp
+++ b/redundant-bmc/src/sibling_impl.cpp
@@ -41,7 +41,7 @@ sdbusplus::async::task<> SiblingImpl::init()
 
     // Attempt to get the D-Bus service name.  It would only
     // be there if the sibling BMC is present.
-    serviceName = co_await getServiceName();
+    serviceName = co_await lookupServiceName();
 
     if (!serviceName.empty())
     {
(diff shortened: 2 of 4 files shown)


redundant-bmc/src/manager.cpp
```diff
@@ -249,17 +249,12 @@ sdbusplus::async::task<std::optional<role_determination::RoleInfo>>
                            RoleReason::systemInventoryNotAvailable};
     }
 
-    // The sibling service must be up and running.
-    if (!providers->getSibling().alive())
+    // If the sibling service isn't on D-Bus, the BMC can't be active.
+    if (!providers->getSibling().alive() &&
```
geissonator: Should this logic be an || ? i.e. if The sibling service is not alive OR the sibling service names is empty we should go to passive?

spinler: The 'not alive' check is asking if the sibling BMC is alive, so just it not being alive doesn't mean this BMC has to be passive.

The important part is the getServiceName, that means that the sibling daemon is broken so we can't trust it to even know if another BMC is running in the system or not.

If it is alive then it is healthy enough to pass this check and not check the service name.  However I just had that in there before so I could save the dbus call if everything was fine.

But now that getServiceName just returns a cached string there probably isn't a need for the alive check so I'll remove it and in hopes of making it less confusing.

spinler: Changed it. Also changed the info trace to an error trace.

geissonator (approved):

redundant-bmc/src/sibling_impl.cpp
```diff
@@ -55,8 +55,7 @@ sdbusplus::async::task<> SiblingImpl::init()
     initialized = true;
 }
 
-// NOLINTNEXTLINE
-sdbusplus::async::task<std::string> SiblingImpl::getServiceName()
+sdbusplus::async::task<std::string> SiblingImpl::lookupServiceName() const
```
RameshIyyar, line 58: > Simplify the check to see if the sibling service is running by checking
if it is providing the State.Decorator.Availability interface on the
'/xyz/openbmc_project/state/bmc1' path,

I think we shouldn’t rely on the Availability interface here. Instead, we should use the Redundancy interface, and the object name should match the SiblingBMC constant value.

Since you’re already accessing the sibling BMC object via one of the hosting interfaces, it should work as-is. However, I believe using the Redundancy interface with SiblingBMC would be the more appropriate approach.

RameshIyyar (approved):

redundant-bmc/src/sibling_impl.cpp
spinler: The Availability interface will always be on D-Bus if rbmc-cfamd  is alive because even if the sibling BMC is missing, the Available property will just be set to false.  that's why I chose that one to use.

That is different than the Redundancy interface, where if the sibling BMC isn't there that won't even be on D-Bus so it can't be used to find the service name.

PR #129, opened 2026-02-18
datetime: Add the necessary targets and service files

Pavithrab7:
The active and passive BMCs must have necessary deamon and settings to be loaded when the targets are triggered. Added the necessary service files and tagged them to the respective targets.

Testing in progress

Signed-off-by: Pavithra Barithaya <pavithrabarithaya07@gmail.com>

### DIFF
--- a/service_files/active-ntp.service
+++ b/service_files/active-ntp.service
@@ -1,0 +1,22 @@
+[Unit]
+Description=Active NTP Service
+
+[Service]
+Type=oneshot
+# Reload settings manager to get the preserved time sync method
+ExecStart=/bin/sh -c 'systemctl reload phosphor-settings-manager.service || true'
+# Wait a moment for settings to be loaded
+ExecStart=/bin/sleep 1
+# Read the saved sync method and apply it
+ExecStart=/bin/sh -c ' \
+    SETTINGS_FILE="/var/lib/phosphor-settings-manager/settings/xyz/openbmc_project/time/sync_method"; \
+    if [ -f "$SETTINGS_FILE" ]; then \
+        SYNC_METHOD=$(grep -oP "(?<=\"TimeSyncMethod\": \")[^\"]*" "$SETTINGS_FILE" 2>/dev/null || echo "xyz.openbmc_project.Time.Synchronization.Method.Manual"); \
+    else \
+        SYNC_METHOD="xyz.openbmc_project.Time.Synchronization.Method.Manual"; \
+    fi; \
+    busctl set-property xyz.openbmc_project.Settings /xyz/openbmc_project/time/sync_method xyz.openbmc_project.Time.Synchronization TimeSyncMethod s "$SYNC_METHOD" \
+'
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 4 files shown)


service_files/active-ntp.service
```diff
@@ -0,0 +1,22 @@
...
+Type=oneshot
+# Reload settings manager to get the preserved time sync method
+ExecStart=/bin/sh -c 'systemctl reload phosphor-settings-manager.service || true'
+# Wait a moment for settings to be loaded
+ExecStart=/bin/sleep 1
+# Read the saved sync method and apply it
+ExecStart=/bin/sh -c ' \
+    SETTINGS_FILE="/var/lib/phosphor-settings-manager/settings/xyz/openbmc_project/time/sync_method"; \
+    if [ -f "$SETTINGS_FILE" ]; then \
+        SYNC_METHOD=$(grep -oP "(?<=\"TimeSyncMethod\": \")[^\"]*" "$SETTINGS_FILE" 2>/dev/null || echo "xyz.openbmc_project.Time.Synchronization.Method.Manual"); \
```
manojkiraneda, line 14: this is pretty complex grepping - and the problem with grepping is that we can miss certain patterns - do we really need this ?

geissonator:
add to approvelist

service_files/active-ntp.service
```diff
@@ -0,0 +1,22 @@
+[Unit]
```
geissonator, line 1: High level I don't think this belongs in phosphor-state-manager and the sleep and complex grep are pretty concerning as well to me. Maybe a design call topic on overall goals and requirements here so we can brainstorm a bit?

target_files/obmc-bmc-active.target
```diff
@@ -1,3 +1,5 @@
 [Unit]
 Description=Active BMC Target
 Conflicts=obmc-bmc-passive.target
+Requires=active-ntp.service chronyd.service
```
geissonator, line 4: You should just use the RBMC bitbake features to ensure the active target requires your new service.

service_files/active-ntp.service
Pavithrab7: yes -- this is needed because during the failover when passive becomes active - when the active ntp service starts it should start with the same mode that earlier active was, so I was trying to grep and see if the file is there and if the mode is there and apply it below.

I can modify it and remove this grepping, can you please tell what could be the better way to do it other than grepping?

service_files/active-ntp.service
Pavithrab7: HI Andrew, sure. 

Let me try to explain on things I added here;
1. Sleep - I added this because the sync method in the settings file is only synced from the rsync manager and in the failover case there needs to be a deamon reload to apply that settings.
I can handle it better yes, need some suggestion.

2. grep - this is needed because during the failover when passive becomes active - when the active ntp service starts it should start with the same mode that earlier active was, so I was trying to grep and see if the file is there and if the mode is there and apply it below.

Sure Andrew, If there is any suggestions for this , please let me know.

I will first brainstorm with my team and later add this topic in the design call sometime next week. 

( I would not be able to discuss it this week as this whole week I would be out evenings for some personal work.)

Pavithrab7: For the place to keep this service file - 
Since I wanted to start these service files with the active and passive targets I thought it would be better to add these in the state manager. Do let me know where I can place these.

Pavithrab7:
Closing this PR as we have addressed this in a different way and we have merged them.

PR #132, opened 2026-02-24
RBMC: Create redundancy related PELs

spinler:
Create PELs for:

- Failover block
- Failover started
- Cannot enable redundancy

### DIFF
--- a/redundant-bmc/src/error_data.cpp
+++ b/redundant-bmc/src/error_data.cpp
@@ -214,4 +214,30 @@ void addDefaultData(const RedundancyInterface& iface, Providers& providers,
     addFWData(providers.getServices(), providers.getSibling(), data);
 }
 
+void addFailoverOptsToData(const FailoverOptions& options,
+                           errors::AdditionalData& data)
+{
+    constexpr auto failoverOptPrefix = "FOOpt:";
+
+    for (const auto& [key, value] : options)
+    {
+        std::string adKey = failoverOptPrefix + key;
+
+        data.emplace(adKey, std::visit(
+                                [](auto&& val) -> std::string {
+                                    using T = std::decay_t<decltype(val)>;
+                                    if constexpr (std::is_same_v<T, bool>)
+                                    {
+                                        return val ? "true" : "false";
+                                    }
+                                    else
+                                    {
+                                        // FailoverOptions can only hold bools
+                                        return "Unsupported option type";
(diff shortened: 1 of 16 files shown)


redundant-bmc/src/error_data.hpp
```diff
@@ -20,4 +20,15 @@ namespace rbmc::errors
 void addDefaultData(const RedundancyInterface& iface, Providers& providers,
                     AdditionalData& data);
 
+using FailoverOptions = std::map<std::string, std::variant<bool>>;
```
aryakpadman: FailoverOptions is also defined in [role_handler.hpp ](https://github.com/spinler/phosphor-state-manager/blob/rbmc_error_logging/redundant-bmc/src/role_handler.hpp#L8)
Will it create conflicts ?

redundant-bmc/src/error_data.cpp
```diff
@@ -214,4 +214,30 @@ void addDefaultData(const RedundancyInterface& iface, Providers& providers,
     addFWData(providers.getServices(), providers.getSibling(), data);
 }
 
+void addFailoverOptsToData(const FailoverOptions& options,
```
aryakpadman, line 217: Commit message : I guess `validateFailoverRequest()` is mistakenly marked as `validateFailureRequest()`

spinler: Good catch,  will fix.

redundant-bmc/src/error_data.hpp
spinler: I couldn't find a good common place to put that. I will see if I can come up with something.

spinler: The smallest header that kinda made sense was errors.hpp, so I moved it there so at least now it is in one place instead of 3.

spinler: I changed my mind I think, I'll just move it into a new header by itself.

RameshIyyar (approved):

redundant-bmc/src/error_data.hpp
spinler: Created a new types.hpp with just this definition for now.  Over time I'd like to move in more things.

aryakpadman: ok..

aryakpadman (approved):

PR #133, opened 2026-02-25
 RBMC: Fill in FailoversNotAllowedReason prop

spinler:
A property to hold the reason that failovers aren't allowed is now on
D-Bus, so fill it in.  It just holds a single value as opposed to
before, because in reality that is all that is useful anyway.

The other commit just simplified rbmc_tool.cpp a bit.

### DIFF
--- a/redundant-bmc/src/redundancy_mgr.cpp
+++ b/redundant-bmc/src/redundancy_mgr.cpp
@@ -13,17 +13,7 @@ RedundancyMgr::RedundancyMgr(sdbusplus::async::context& ctx,
                              Providers& providers, RedundancyInterface& iface) :
     ctx(ctx), providers(providers), redundancyInterface(iface),
     manualDisable(iface.disable_redundancy_override())
-{
-    try
-    {
-        data::remove(data::key::failoversNotAllowedReasons);
-    }
-    catch (const std::exception& e)
-    {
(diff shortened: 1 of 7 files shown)


aryakpadman (approved):

redundant-bmc/src/redundancy_mgr.cpp
```diff
@@ -265,43 +255,25 @@ void RedundancyMgr::determineAndSetFailoversAllowed()
         .failoverInProgress = failoverInProgress,
         .systemState = systemState.value_or(SystemState::other)};
 
-    auto notAllowedReasons = fona::getFailoversNotAllowedReasons(input);
+    auto reason = fona::getFailoversNotAllowedReason(input);
 
-    // TODO: Put the reasons on D-Bus. For now, just save the
-    // descriptions in the normal spot for rbmctool.
+    redundancyInterface.failovers_not_allowed_reason(reason);
```
RameshIyyar, line 260: is not persisted?

spinler: Correct, it'll just be on D-Bus now since it will just be re-calculated every time and it will simplify code.  I would also like to do that with  the RedundancyDisabledReasons to make things simpler.  rbmctool already looks at d-bus, and that'll run during a dump.

RameshIyyar (approved):

PR #134, opened 2026-03-04
RBMC: Don't start already started targets

spinler:
Starting a unit that is already active will still run all of the services that depend on it which isn't desired, so before starting the active or passive targets check if they are already active and don't do anything if they are.

The only case this comes into play is if this service is restarted after the target has already been started.

Tested:
- First time: service still starts
- Already running: see:

```
Unit obmc-bmc-active.target is already active, not starting again
```

Change-Id: Ie5002d21e31d481f6486ab7eda005a1c8a6277fd

### DIFF
--- a/redundant-bmc/src/services_impl.cpp
+++ b/redundant-bmc/src/services_impl.cpp
@@ -527,15 +527,33 @@ sdbusplus::async::task<> ServicesImpl::startUnit(
     const std::string& unitName, std::chrono::seconds timeout) const
 {
     using namespace std::chrono_literals;
+
+    auto currentState = co_await getUnitState(unitName);
+    if (currentState == "active")
+    {
+        lg2::info("Unit {UNIT} is already active, not starting again", "UNIT",
+                  unitName);
+        co_return;
+    }
+
     constexpr auto systemd = sdbusplus::async::proxy()
                                  .service(service::systemd)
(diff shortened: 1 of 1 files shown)


redundant-bmc/src/services_impl.cpp
```diff
@@ -527,6 +527,14 @@ sdbusplus::async::task<> ServicesImpl::startUnit(
     const std::string& unitName, std::chrono::seconds timeout) const
 {
     using namespace std::chrono_literals;
+
+    if (co_await getUnitState(unitName) == "active")
```
RameshIyyar: Interesting, does it restart depend services if its running already?

spinler: It's things like this I think:
```
[Service]
ExecStart=/usr/bin/wait-vpd-parser
Type=oneshot  <---------------------------

```

geissonator: Yeah I've def had some fun with this type of issue over the years. The only other thing I'd add is if you have any concern it could be in the process of starting you'd want to add a check for "activating" as well.

spinler: Good idea, I'll add that.

aryakpadman (approved):

geissonator (approved):

PR #135, opened 2026-03-04
RBMC: Keep track of in-progress waits

spinler:
Introduce wait tracker APIs so that in all the places the code is waiting for something to happen what is being waited on can be displayed in rbmctool.

  - Added new wait tracker APIs to monitor what the code is currently waiting on during various state transitions
  - Wait states are persisted to the data file so rbmctool can read them
  - Call the new APIs in all of the relevant places.
  - Create a persistent_data test fixture to make handling the data file easier in unit tests.

### DIFF
--- a/redundant-bmc/src/wait_tracker.cpp
+++ b/redundant-bmc/src/wait_tracker.cpp
@@ -1,0 +1,125 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#include "wait_tracker.hpp"
+
+#include "persistent_data.hpp"
+
+#include <phosphor-logging/lg2.hpp>
+
+#include <algorithm>
+
+namespace rbmc
+{
+
+std::string waitToString(Wait wait)
+{
+    switch (wait)
+    {
+        case Wait::startUnit:
+            return "StartUnit";
+        case Wait::systemInventoryPath:
+            return "SystemInventoryPath";
+        case Wait::systemInventoryStatus:
+            return "SystemInventoryStatus";
+        case Wait::peerConnection:
+            return "PeerConnection";
+        case Wait::siblingAlive:
+            return "SiblingAlive";
+        case Wait::siblingRole:
+            return "SiblingRole";
+        case Wait::siblingBMCSteadyState:
+            return "SiblingBMCSteadyState";
+        case Wait::siblingHealth:
+            return "SiblingHealth";
+        case Wait::fullSync:
+            return "FullSync";
+        case Wait::failoverImminent:
+            return "FailoverImminent";
+    }
+    return "Unknown";
+}
+
+std::vector<std::string> getTrackedWaits()
+{
+    try
+    {
+        using U = std::underlying_type_t<Wait>;
+        auto waits = data::read<std::vector<U>>(data::key::trackedWaits)
+                         .value_or(std::vector<U>{});
+
+        std::vector<std::string> result;
+        result.reserve(waits.size());
+        for (auto w : waits)
+        {
+            result.push_back(waitToString(static_cast<Wait>(w)));
+        }
+        return result;
+    }
+    catch (const std::exception& e)
+    {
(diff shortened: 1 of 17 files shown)


geissonator (commented):
Hmm, not sure how I feel about this one. I see the usefulness, but I also see a lot of extra code and complexity that's now running with no way to disable it. Was a debug journal indicating where the code is waiting not good enough? Is there a way we could at least compile it out in production?

spinler:
> Hmm, not sure how I feel about this one. I see the usefulness, but I also see a lot of extra code and complexity that's now running with no way to disable it. Was a debug journal indicating where the code is waiting not good enough? Is there a way we could at least compile it out in production?

It's all just a fancy wrapper around writing an entry in the data.json.

The issue is that if someone is going to be looking in the journal to figure out what's going on, that person is always going to be me because nobody else, especially non BMC developers will be able to.  But if I just make it obvious what the code is waiting on, especially that 6-minute wait before a role is even determined, then there will be no question by anyone when they use rbmctool.

It's definitely also great to use for my development.

redundant-bmc/src/wait_tracker.cpp
```diff
@@ -0,0 +1,125 @@
...
+    try
+    {
+        using U = std::underlying_type_t<Wait>;
+        auto waits = data::read<std::vector<U>>(data::key::trackedWaits)
+                         .value_or(std::vector<U>{});
+
+        if (!std::ranges::contains(waits, std::to_underlying(wait)))
+        {
+            waits.push_back(std::to_underlying(wait));
+            data::write(data::key::trackedWaits, waits);
```
aryakpadman, line 76: Just a thought, do we really want to write into file also ? Can't we keep it in-memory alone? Also these waits are a transition states only.

spinler: rbmctool needs a way to be able to get them.

aryakpadman: Oh my bad.. Thanks!

aryakpadman (approved):

geissonator:
> > Hmm, not sure how I feel about this one. I see the usefulness, but I also see a lot of extra code and complexity that's now running with no way to disable it. Was a debug journal indicating where the code is waiting not good enough? Is there a way we could at least compile it out in production?
> 
> It's all just a fancy wrapper around writing an entry in the data.json.
> 
> The issue is that if someone is going to be looking in the journal to figure out what's going on, that person is always going to be me because nobody else, especially non BMC developers will be able to. But if I just make it obvious what the code is waiting on, especially that 6-minute wait before a role is even determined, then there will be no question by anyone when they use rbmctool.
> 
> It's definitely also great to use for my development.

Here's my issues. I'm not gonna die on this hill though since it's your app and you get to maintain it so do what you think is best :) 


1. Excessive Production Code Complexity for Debug Info

Adds ~600 lines of code (+606/-83 across 17 files)
Introduces a complete wait tracking infrastructure with persistent storage
Every wait point now requires explicit tracking calls scattered throughout the codebase
The ScopeWaitTracker RAII wrapper adds overhead to every wait operation

2. Tight Coupling of Debug Infrastructure

Wait tracking is deeply embedded in production code paths (active_role_handler, redundancy_mgr, timer, etc.)
Persistent data file is modified on every wait state change
Debug functionality cannot be easily disabled or compiled out

3. Design Concerns

Persistent Storage for Transient State: Writing temporary "what am I waiting for" state to disk seems excessive. This is runtime debug info, not data that needs to survive reboots.

Maintenance Burden: Every new wait point requires manual instrumentation

spinler:
that is quite the writeup...

My next PR https://github.com/ibm-openbmc/phosphor-state-manager/pull/137 is more important and unfortunately probably bigger.  I can let this one sit for now.

spinler:
Still need some functionality like this, but will do slightly differently.

PR #137, opened 2026-03-16
RBMC: Start using PeerConnected property to know inter-BMC network status

spinler:
This PR adds code to start watching the PeerConnected property that says if the network connection is good between the BMCs. 

It will wait up to 10 minutes for the property to go to Connected continuing on and trying to enable redundancy.  At that point if it isn't connected, redundancy will be disabled with the reason of 'NetworkError'.

It can handle the connection being lost and regained at steady-state.

### DIFF
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -82,6 +82,10 @@ will wait up to ten minutes total for the passive BMC to get to a steady state
 (assuming the passive BMC is alive), which would either be `Ready` or
 `Quiesced`.
 
+There is also a ten minute wait for the network connection between the BMCs.
+This is in parallel with the steady state wait so if one is complete first it
+will still wait for the other.
+
 After the passive BMC reaches steady state, it will then check the following
 items to see if redundancy can be enabled:
 
@@ -130,6 +134,21 @@ wasn't running.
 Note that redundancy cannot be enabled at runtime if the system wasn't booted
 with redundancy enabled. A concurrent maintenance operation would be necessary
 in that case.
+
+### Passive BMC loses network connection
+
+A loss of the network connection between the BMCs will start a five minute timer
+similar to how it was done for a heartbeat loss. At five minutes, redundancy
+will be disabled and an event log will be created. If it comes back before then,
+it will go through the code to calculate redundancy followed by a full sync to
+sync over any files that might have changed when the network was down.
+
+The heartbeat timer will take precedence over this one. So if the heartbeat
+isn't currently active when the network loss occurs, the new timer won't be
+started since the overall loss of the passive BMC is already being handled.
+
+In addition, if the network loss timer is already running when the heartbeat
+loss is noticed, it will be canceled when the heartbeat loss timer is started.
 
 ## Interacting with Data Sync on the Active BMC
 
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -94,6 +94,7 @@ items to see if redundancy can be enabled:
    so.
 1. The sibling BMC has been provisioned.
 1. The firmware versions are the same on the BMCs.
+1. The network between the BMCs is connected.
 1. If attempting to enable any time at runtime, redundancy must have been
    enabled when runtime was first reached.
 
--- a/redundant-bmc/src/active_role_handler.cpp
+++ b/redundant-bmc/src/active_role_handler.cpp
@@ -50,11 +50,13 @@ sdbusplus::async::task<> ActiveRoleHandler::start()
 
     if (sibling.alive())
     {
-        // Make sure the sibling had time to get its role assigned, and
-        // also wait for it to hit steady state as redundancy can only
-        // be enabled if the sibling BMC is at the Ready state.
+        // Before trying to enable redundancy, wait for:
+        // 1. Sibling to have its role assigned.
+        // 2. Sibling to hit steady state (Ready needed for redundancy)
+        // 3. The network to connect to the sibling BMC.
         co_await sdbusplus::async::execution::when_all(
-            sibling.waitForSiblingRole(), sibling.waitForBMCSteadyState());
+            sibling.waitForSiblingRole(), sibling.waitForBMCSteadyState(),
+            services.waitForPeerConnection());
     }
 
     co_await redMgr.determineRedundancyAndSync();
@@ -109,8 +111,10 @@ sdbusplus::async::task<> ActiveRoleHandler::siblingHealthy()
     stopSiblingWatches();
 
     auto& sibling = providers.getSibling();
+    auto& services = providers.getServices();
     co_await sdbusplus::async::execution::when_all(
-        sibling.waitForSiblingRole(), sibling.waitForBMCSteadyState());
+        sibling.waitForSiblingRole(), sibling.waitForBMCSteadyState(),
+        services.waitForPeerConnection());
 
     lg2::info("Attempting to enable redundancy now that sibling is back");
     providers.getSyncInterface().clearFullSyncComplete();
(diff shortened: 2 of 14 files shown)


redundant-bmc/src/meson.build
```diff
@@ -20,6 +20,7 @@ executable(
     'error_data.cpp',
     'manager.cpp',
     'passive_role_handler.cpp',
+    'peer_connection_state.cpp',
```
geissonator: The commit message that added this looks a little off:
```
It can have these values:
 - NotDetermined
 - InProgress
 - NotConnected
 - InProgress
```
Should that last one be "Connected"?

Also I'm still a bit confused on the need for the "final" part of this. Why can't we just wait until we see "Connected"? I guess it's more about the error paths where it is Connected and then goes NotConnected? So we'd wait a bit for Connected to come back and at some point set final=true if it stays NotConnected and disable redundancy?

redundant-bmc/src/services.hpp
```diff
@@ -3,6 +3,7 @@
 #include "errors.hpp"
 
 #include <sdbusplus/async.hpp>
+#include <xyz/openbmc_project/Provisioning/Provisioning/common.hpp>
```
geissonator: Oh boy, I see this in the commit msg as well, do we need to change all of this over to Paired now? Probably best to just get all the basics in place and then go back and do one big commit to move everything over.

redundant-bmc/src/active_role_handler.cpp
```diff
@@ -65,7 +65,7 @@ sdbusplus::async::task<> ActiveRoleHandler::start()
 
     startSyncHealthWatch();
 
-    co_return;
+    startPeerConnectedWatch();
```
geissonator, line 68: The test in this commit msg confused me a bit:

PeerConnection is only briefly lost:
```
phosphor-rbmc-state-manager[2041]: PeerConnected changed to False
phosphor-rbmc-state-manager[2041]: Disabling redundancy in 5 minutes if peer connection doesn't come back
phosphor-rbmc-state-manager[2041]: PeerConnected changed to True
phosphor-rbmc-state-manager[2041]: Done waiting for sibling steady state. State = xyz.openbmc_project.State.BMC.BMCState.Ready
phosphor-rbmc-state-manager[2041]: waitForPeerConnection initial value = xyz.openbmc_project.Provisioning.Provisioning.PeerConnectionStatus.Connected, isFinal = True
phosphor-rbmc-state-manager[2041]: Attempting to enable redundancy now that sibling is back
phosphor-rbmc-state-manager[2041]: Enabling redundancy
```

Those last 2 messages imply redundancy was disabled? But it should have only been disabled if we hit the 5 minute timer?

geissonator (commented):
Overall looks good to me. I'm still annoyed with github that we can't just comment on commit messages.

redundant-bmc/src/meson.build
spinler: Thanks, that should indeed be Connected.  Will fix.

The problem is when we enter the wait loop but it is still at NotConnected.  So then I have to wait for it to change to something else, and then change back again to either Connected or NotConnected.  But it's also possible it would be at NotConnected and not change again, so it would have to wait the whole time unnecessarily.

So I put the complexity of that into a separate class.  

The 'isFinal' basically means the value is at it's final value and we don't need to wait for it to change (i.e. go NotConnected->InProgress->(Connected|NotConnected).

redundant-bmc/src/services.hpp
spinler: Yea.  if the timing would have aligned better I could have just used Pairing in the first place, but oh well.

redundant-bmc/src/active_role_handler.cpp
spinler: Oh, those just print every time the code runs through the redundancy analysis, like:
```
    lg2::info("Attempting to enable redundancy now that sibling is back");
    providers.getSyncInterface().clearFullSyncComplete();
    co_await redMgr.determineRedundancyAndSync();
```

I can try to clean up those traces (in a future commit).

spinler:
After doing some failover stress testing, I'm seeing erratic behavior of PeerConnected, so I may need to make changes.

So can hold off on further reviews for now.  I will make an update when it's all figured out.

spinler:
Yea, after a lot of  testing, I have to give up on trying to be clever, and I will just wait 5 minutes for Connected, no matter what it starts at.

redundant-bmc/src/meson.build
spinler: Going away from t his.

spinler:
Simplified things.

spinler:
Just saw it take 6 minutes in the lab.  Will up the timeout to 10 I guess.

redundant-bmc/README.md
```diff
@@ -82,6 +82,10 @@ will wait up to ten minutes total for the passive BMC to get to a steady state
 (assuming the passive BMC is alive), which would either be `Ready` or
 `Quiesced`.
 
+There is also a two minute wait for the network connection between the BMCs.
```
spinler: I bumped the wait up to 10 so I need to update this.

redundant-bmc/src/redundancy_mgr.cpp
```diff
@@ -203,7 +203,7 @@ void RedundancyMgr::initSystemState()
     auto& services = providers.getServices();
 
     services.addSystemStateCallback(
-        std::bind_front(&RedundancyMgr::systemStateChange, this));
+        Role::Active, std::bind_front(&RedundancyMgr::systemStateChange, this));
```
RameshIyyar, line 206: I’m not sure whether I fully convinced by the idea of “adding role-specific logic in the redundancy manager class.” I would lean more toward an Active/Passive Role Handler approach.

I understand that the system (host) state can be obtained from the Active BMC—so wouldn’t it make more sense to handle this within the role handler class itself?

redundant-bmc/src/passive_role_handler.cpp
```diff
@@ -350,4 +353,13 @@ auto PassiveRoleHandler::getFailoverBlockedReason(
     co_return fo_blocked::getFailoverBlockedReason(input);
 }
 
+void PassiveRoleHandler::peerConnectedChange([[maybe_unused]] bool connected)
+{
+    // If connected and all other conditions succeed for a full sync,
+    // will start one, otherwise will stop background syncing.
+    // If the active BMC just came back from a reboot, most likely
```
RameshIyyar, line 360: I believe, we will initiate a full sync when the sibling (active) enabling redundancy since that would be mirrored in the Passive.

(Thats what I remembered from our previous discussion)

redundant-bmc/src/active_role_handler.cpp
```diff
@@ -93,6 +93,9 @@ void ActiveRoleHandler::siblingHealthChange(bool alive)
                 "Disabling redundancy in {TIME} minutes if sibling doesn't come back",
                 "TIME", siblingHealthTimeout.count());
             siblingHealthTimer.start(siblingHealthTimeout);
+
+            // Background sync won't work without a healthy sibling
```
RameshIyyar, line 97: Why?

After timer, we will disable redundancy that will take care to disable sync, right? - I believe, you are covering CFAMs communication issues, right?

spinler: In my testing I saw a background sync fail when the passive BMC was rebooted because it was left on. That may have even caused a new PEL though I don't remember for sure.

At this point we know something is wrong with the passive BMC, so I think it makes sense to turn of background sync, as a full sync will be done anyway when it comes back again.

redundant-bmc/src/passive_role_handler.cpp
spinler: Yea, that's right.

redundant-bmc/src/redundancy_mgr.cpp
spinler: That role is just a key into the callback map in the services class:
    /**
     * @brief The functions to call when the system state changes
     */
    std::map<Role, SystemStateCallback> systemStateCBs;

I don't feel strongly about what the key is, though I'd still like to keep it a map so it's easy to just unregister a single callback.

Another option is just a string, which probably makes more sense anyway.  The main thing is I'd like all the 2 callbacks in services.hpp and 5 in sibling.hpp to all be the same.

Does that sound better?  if so, i can do a separate PR to convert them over.

The RedundancyMgr class only exists as part of the ActiveRoleHandler, so it doesn't really do any checking between roles.

I could probably move the system state callbacks into ActiveRoleHandler, but I'd like to keep the low-level handling in RedMgr so it could just call the existing SystemStateChange function there on it I guess.

RameshIyyar: Ok, I am fine with current style also, no harm in that.

redundant-bmc/src/active_role_handler.cpp
RameshIyyar: Ok make sense.

RameshIyyar (approved):

redundant-bmc/src/services_impl.cpp
```diff
@@ -830,4 +928,42 @@ sdbusplus::async::task<> ServicesImpl::logError(
     }
 }
 
+sdbusplus::async::task<> ServicesImpl::waitForPeerConnection()
+{
+    using namespace std::chrono_literals;
+    std::chrono::minutes timeout{10};
```
geissonator, line 934: I think there's an argument to be had to make this a configurable meson option, or at least put some more comments around here to indicate this is where this is set as it's a very important default.

redundant-bmc/src/active_role_handler.cpp
```diff
@@ -50,11 +50,13 @@ sdbusplus::async::task<> ActiveRoleHandler::start()
 
     if (sibling.alive())
     {
-        // Make sure the sibling had time to get its role assigned, and
-        // also wait for it to hit steady state as redundancy can only
-        // be enabled if the sibling BMC is at the Ready state.
+        // Before trying to enable redundancy, wait for:
```
geissonator, line 53: nit: commit message test still shows 5 minutes as timeout

geissonator (approved):

PR #138, opened 2026-04-03
RBMC: Sync handling improvements

spinler:
1. Combine all of ActiveRoleHandler's start/stop watch calls into single stop and start all watches functions.  This also fixes a situation where the sync health handler got called while a full sync was running and failed.  It doesn't need to be called as the code running the full sync already handles the fail.
2. As soon as the passive BMC health comes back after it was offline, clear the full sync complete flag to turn off failovers allowed while the code is waiting for the passive to come all the way back until the full sync is done.

The 2nd one is because now with the PeerConnected check, the code can be waiting for a lot longer before it even gets to the full sync again, so it should be obvious this isn't failovers allowed steady state.

### DIFF
--- a/redundant-bmc/src/active_role_handler.cpp
+++ b/redundant-bmc/src/active_role_handler.cpp
@@ -123,6 +123,12 @@ sdbusplus::async::task<> ActiveRoleHandler::siblingHealthy()
     auto& sibling = providers.getSibling();
     auto& services = providers.getServices();
 
+    // As redundancy could be enabled here if passive
+    // wasn't gone long, hold off failovers until the
+    // full sync is done.
+    providers.getSyncInterface().clearFullSyncComplete();
+    redMgr.determineAndSetFailoversAllowed();
+
     // Before trying to enable redundancy, wait for:
     // 1. Sibling to have its role assigned.
     // 2. Sibling to hit steady state (Ready needed for redundancy)
@@ -132,7 +138,6 @@ sdbusplus::async::task<> ActiveRoleHandler::siblingHealthy()
         services.waitForPeerConnection());
 
     lg2::info("Attempting to enable redundancy now that sibling is back");
-    providers.getSyncInterface().clearFullSyncComplete();
     co_await redMgr.determineRedundancyAndSync();
 
     startSiblingWatches();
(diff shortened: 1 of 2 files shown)


aryakpadman (approved):

redundant-bmc/src/active_role_handler.cpp
```diff
@@ -123,6 +123,12 @@ sdbusplus::async::task<> ActiveRoleHandler::siblingHealthy()
     auto& sibling = providers.getSibling();
     auto& services = providers.getServices();
 
+    // As redundancy could be enabled here if passive
+    // wasn't gone long, hold off failovers until the
+    // full sync is done.
+    providers.getSyncInterface().clearFullSyncComplete();
+    redMgr.determineAndSetFailoversAllowed();
```
RameshIyyar, line 130: We shouldn’t allow failover once we detect that the sibling has entered a reboot state, right?

What is the reason for not allowing the failover during the window after reboot until the sibling reaches a healthy/ready state?

spinler: PeerConnected is the big problem, now that it can take anywhere from seconds to minutes for that to go to connected.

RameshIyyar: Hmm, I feel we could still state that failover (FO) is not allowed once the passive BMC reboot begins, though it’s not a major concern.

I’m thinking of the scenario where redundancy isn’t disabled, so FO is technically allowed—but since the passive BMC is unavailable, no one can actually initiate a failover from the Active BMC, HMC, or PHYP.

RameshIyyar: and I think even if we handled this case does not help much because it wont be visible to the P-BMC..I think we could ignore now.

RameshIyyar (approved):

redundant-bmc/src/active_role_handler.cpp
spinler: In this case:
> I’m thinking of the scenario where redundancy isn’t disabled, so FO is technically allowed—but since the passive BMC is unavailable, no one can actually initiate a failover from the Active BMC, HMC, or PHYP.

We had talked about this in one of our interlock meetings before, and decided with Dean that since the passive BMC isn't online that we don't need to turn off failovers allowed.

I suspect when people start actually looking at all of these properties to know when to do failovers things like this may need additional discussion.

PR #143, opened 2026-04-14
RBMC:  Add unit tests for Manager

spinler:
Start adding test cases for the Manager class, which involve choosing
roles, enabling redundancy, and failing over.

With how many different pieces of data are used when doing these
operations, the variations are almost endless.  This commit adds a few
to start with.

There are lots of function calls in the mocked classes, so to keep the
test cases at a reasonable length only the important functions for each
specific test case have an EXPECT_CALL.

The test case framework will wait for the  EventRecorder's events to
occur instead of just sleeping for a hardcoded amount of time in
order to make the tests as fast as possible.

Other commits:
- Add an internal event recorder
- Create a static library to share between the daemon, rbmctool, and tests
- Add the mocks for all of the classes in Providers
- Testability improvements for the persistent data functions
- Fix a destruction ordering bug in Manager
- Make the heartbeat interval be configurable to speed up the testcases


redundant-bmc/src/manager.hpp
```diff
@@ -178,6 +225,16 @@ class Manager :
...
+
+    /**
+     * @brief The interval between heartbeats
+     */
+    std::chrono::milliseconds heartbeatInterval;
+
+    /**
+     * @brief Set of checkpoints that have been reached
+     */
+    std::set<Checkpoint> checkpoints;
```
spinler: I think I should move this out of Manager into somewhere like Providers, and probably make it its own class, so  other code can access it easier.

spinler: Done

redundant-bmc/test/manager_test.cpp
```diff
@@ -0,0 +1,864 @@
...
+        // If timed out, stop regardless
+        ctx.request_stop();
+    }
+
+    /**
+     * @brief Test helper for waiting for a successful failover
+     *
+     * A standalone function so it has the proper lifetime
+     */
+    sdbusplus::async::task<> runStartFailoverTest()
```
spinler: I found a way to run this code inline that doesn't make the sanitzers mad. Will change it.

spinler:
I need some of these commits in another PR. Also this has conflicts.
Will close this and starting putting up commits separately, should make things easier to review.

PR #146, opened 2026-04-22
RBMC Pairing monitoring enhancements

spinler:
1. Don't use the sibling pairing value in role determination since it will aleady be passive anyway.
2. When a BMC is paired, clear the 'cannot be active' indication so that it can become active in the future without a reboot.
3. The sibling pairing result is now available in the sibling daemon so get it from there.

The self pairing value is still hardcoded to true because in simulation there is no pairing done yet.

### DIFF
--- a/redundant-bmc/src/active_role_handler.cpp
+++ b/redundant-bmc/src/active_role_handler.cpp
@@ -57,6 +57,15 @@ sdbusplus::async::task<> ActiveRoleHandler::start()
         co_await sdbusplus::async::execution::when_all(
             sibling.waitForSiblingRole(), sibling.waitForBMCSteadyState(),
             services.waitForPeerConnection());
+
+        // If PeerConnected == true and sibling provisioned == false
+        // a small delay will be needed to let the new provisioned
+        // value propagate to this BMC.
+        if (services.getPeerConnected() &&
+            !sibling.getProvisioned().value_or(true))
+        {
+            co_await sibling.pauseForDataPropagation();
+        }
     }
 
     co_await redMgr.determineRedundancyAndSync();
@@ -131,6 +140,13 @@ sdbusplus::async::task<> ActiveRoleHandler::siblingHealthy()
     co_await sdbusplus::async::execution::when_all(
         sibling.waitForSiblingRole(), sibling.waitForBMCSteadyState(),
         services.waitForPeerConnection());
+
+    // Just like in start(), a delay may be needed to let
+    // the sibling provisioned value propagate.
+    if (services.getPeerConnected() && !sibling.getProvisioned().value_or(true))
+    {
+        co_await sibling.pauseForDataPropagation();
+    }
 
     lg2::info("Attempting to enable redundancy now that sibling is back");
     co_await redMgr.determineRedundancyAndSync();
--- a/redundant-bmc/src/manager.cpp
+++ b/redundant-bmc/src/manager.cpp
@@ -105,6 +105,11 @@ sdbusplus::async::task<> Manager::startup()
     co_await postStartupClearFOInProgress();
 
     spawnRoleHandler();
+
+    if (!services.getProvisioned())
+    {
+        setupProvisionedWatch();
+    }
 }
 
 // NOLINTNEXTLINE
@@ -453,6 +458,61 @@ sdbusplus::async::task<> Manager::doFailoverFromPassive(Requester requester)
     redundancyInterface.failover_in_progress(false);
 
     co_await active->failoverDetermineRedundancy();
+}
+
+void Manager::setupProvisionedWatch()
+{
+    providers->getServices().addProvisionedCallback(
+        Role::Passive,
+        std::bind_front(&Manager::provisionedChangeHandler, this));
+}
+
+void Manager::provisionedChangeHandler(bool provisioned)
+{
+    ctx.spawn(handleProvisionedChange(provisioned));
+}
+
+sdbusplus::async::task<> Manager::handleProvisionedChange(bool provisioned)
+{
+    if (!provisioned)
+    {
+        co_return;
+    }
+
+    // Doublecheck the BMC is passive.  It should be.
+    if (redundancyInterface.role() != Role::Passive)
+    {
+        lg2::warning(
+            "Provisioned just changed to true but BMC not already passive?");
+        providers->getServices().removeProvisionedCallback(Role::Passive);
+        co_return;
+    }
+
+    // If being passive isn't still required, clear the
+    // reasons_for_no_redundancy property which will let the other BMC
+    // know this BMC can be active again.
+
+    auto passiveRoleInfo = co_await determinePassiveRoleIfRequired();
+
+    if (!passiveRoleInfo.has_value())
+    {
+        lg2::info("This BMC is no longer required to be passive");
+
+        // Note: Leave chosePassiveDueToError so the next reboot still sees it.
+
+        redundancyInterface.reasons_for_no_redundancy({});
+
+        // No need for future callbacks.
+        providers->getServices().removeProvisionedCallback(Role::Passive);
+    }
+    else
+    {
+        lg2::warning(
+            "After provisioned property change to true, BMC still required to be passive: {REASON}",
+            "REASON",
+            role_determination::getRoleReasonDescription(
+                passiveRoleInfo->reason));
+    }
 }
 
 } // namespace rbmc
--- a/redundant-bmc/src/manager.cpp
+++ b/redundant-bmc/src/manager.cpp
@@ -189,7 +189,6 @@ sdbusplus::async::task<role_determination::RoleInfo> Manager::determineRole()
         // Note:  If these returned nullopts, the algorithm wouldn't use
         //        them anyway because there would be no heartbeat.
         auto siblingRole = sibling.getRole().value_or(Role::Unknown);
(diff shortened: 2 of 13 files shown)


redundant-bmc/src/active_role_handler.cpp
```diff
@@ -57,6 +57,15 @@ sdbusplus::async::task<> ActiveRoleHandler::start()
...
             sibling.waitForSiblingRole(), sibling.waitForBMCSteadyState(),
             services.waitForPeerConnection());
+
+        // If PeerConnected == true and sibling provisioned == false
+        // a small delay will be needed to let the new provisioned
+        // value propagate to this BMC.
+        if (services.getPeerConnected() &&
+            !sibling.getProvisioned().value_or(true))
+        {
+            co_await sibling.pauseForDataPropagation();
```
aryakpadman, line 67: do we need to check the Provisioned value after delay, instead of blindly going ahead ?

spinler: There really isn't much that can be done, so I don't think so.  If it isn't true by the time the delay is done then it never would be and redundancy would be disabled.

aryakpadman (approved):

redundant-bmc/src/role_determination.cpp
```diff
@@ -21,11 +21,6 @@ RoleInfo determineRole(const Input& input)
         result = {Role::Active, RoleReason::siblingNotAlive};
     }
 
-    else if (!input.siblingProvisioned)
```
RameshIyyar, line 24: I believe, we add a condition in the future to become this BMC as Passive if this BMC is not provisioned, right?

redundant-bmc/src/manager.cpp
```diff
@@ -455,4 +460,59 @@ sdbusplus::async::task<> Manager::doFailoverFromPassive(Requester requester)
...
+        // Note: Leave chosePassiveDueToError so the next reboot still sees it.
+
+        redundancyInterface.reasons_for_no_redundancy({});
+
+        // No need for future callbacks.
+        providers->getServices().removeProvisionedCallback(Role::Passive);
+    }
+    else
+    {
+        lg2::info(
```
RameshIyyar: warning?

redundant-bmc/src/services_impl.cpp
RameshIyyar, line 711: Is it required to change now?

redundant-bmc/src/active_role_handler.cpp
```diff
@@ -57,6 +57,15 @@ sdbusplus::async::task<> ActiveRoleHandler::start()
         co_await sdbusplus::async::execution::when_all(
             sibling.waitForSiblingRole(), sibling.waitForBMCSteadyState(),
             services.waitForPeerConnection());
+
```
RameshIyyar, line 60: > A BMC still can't go from passive to active without a reboot or
failover, so this won't prevent a reboot in the case of two
unprovisioned BMCs.

Hmm...Both BMCs initially come up in an unpaired, passive state with redundancy disabled. Self-pairing begins on BMC0, and the pair-app on BMC0 updates the provisioned/paired property. This is then observed by the rbmc manager, which indicates that a provisioned BMC can transition to Active by either reboot or failover...

But who is responsible for triggering the reboot to make one BMC0 Active?

In the scenario where both BMCs start in an unpaired state and BMC0 completes pairing, it still remains Passive in the MFG/Lab/Simics setup—so who initiates the transition (via reboot) for it to become Active?

I’m exploring whether we can avoid a reboot. What if the rbmc manager change the paired BMC to Active and starts the Active target when the sibling BMC is also in a Passive state? Redundancy would remain disabled and could be re-enabled once the peer BMC pairing is completed from the Active BMC. The Peer BMC pairing should start after the active target is completed (or I think we no need to wait too)...

redundant-bmc/src/rbmc_tool.cpp
```diff
@@ -294,11 +294,19 @@ sdbusplus::async::task<> getSiblingBMCInfo(sdbusplus::async::context& ctx,
...
+                             .service(siblingService)
+                             .path(path.str)
+                             .properties();
+
         output["Redundancy Enabled"] = rProps.redundancy_enabled;
         output["Failovers Allowed"] = rProps.failovers_allowed;
         output["BMC State"] = getPDIEnumString(state);
         output["FW Version Hash"] = fwVersion;
-        output["Paired"] = true; // TODO
+        if (!provProps.provisioned)
```
RameshIyyar, line 306: I think its not correct, 

We need to rely on the PeerProvisioned Signal, right? https://github.com/ibm-openbmc/phosphor-dbus-interfaces/blob/1210/yaml/xyz/openbmc_project/Provisioning/Provisioning.interface.yaml#L35

redundant-bmc/src/sibling_impl.cpp
```diff
@@ -19,6 +20,8 @@ using VersionIntf = sdbusplus::common::xyz::openbmc_project::software::Version;
 using BMCStateIntf = sdbusplus::common::xyz::openbmc_project::state::BMC;
 using AvailIntf =
     sdbusplus::common::xyz::openbmc_project::state::decorator::Availability;
+using ProvisioningIntf =
```
RameshIyyar, line 23: Commit Message:

> the sibling daemon on the provisioning interface on the state/bmc1 path.

I’m a bit confused—are you referring to the pairing app? If so, we should be checking the PeerProvisioned signal to determine the sibling BMC’s provision state.

Also, I’m trying to understand where you plan to use the sibling provision state. In the initial commit, the logic that considered the sibling’s provision state while determining the role was removed—so is there another place where this check is still needed?

redundant-bmc/src/active_role_handler.cpp
spinler: I'm still pushing back on having to create a new code path that goes from passive to active without a failover.  for the following reasons:

1. It's a new code path that would need to be well tested on real hardware, but in the lab it would hardly ever be tested.
2. I can't stop someone from doing a self-pair on both BMCs at the same time, which would lead to dual active BMCs.

I consider it beyond the scope of this commit anyway.

redundant-bmc/src/manager.cpp
spinler: good idea, will fix.

redundant-bmc/src/rbmc_tool.cpp
spinler: I'm not sure how rbmctool.cpp would use a signal?

When the peer is paired/provisioned, the 'provisioned' property on the peer goes to true, which my commit https://github.ibm.com/openbmc/openpower-proc-control/commit/da22decf6bde2265f4e2cf3db465717b251294ab now puts in the local CFAM and then reads from the sibling CFAM and puts it on dbus using the Provisioning interface.  That's what's being read here.

redundant-bmc/src/role_determination.cpp
spinler: Yea, kind of.  It doesn't even make it into this function if the BMC isn't provisioned. It hits manager::determinePassiveRoleIfRequired before the heartbeat even starts that sets the role.

redundant-bmc/src/services_impl.cpp
spinler: I wish, but there's still no way to do the self pair in simics so the BMC would go to passive. It works fine on rbmc-prototype though.

These commits are dealing with the pairing status of the sibling, which is accurate because rbmc-cfamd is watching the actual Provisioning interface on the other BMC.

redundant-bmc/src/sibling_impl.cpp
spinler: I had actually forgotten that signal existed, but I don't think I could have used it anyway because what if my code isn't running at the time so I missed it?  Also, then I would have to persist it myself to know it across reboots.

How i'm doing it instead is using the 'Paired/Provisioned' bit we already have in the CFAM.  Each BMC watches its own Provisioned property in provisioningd and writes the value to the CFAM, and then on the other side puts it on DBus again on state/bmc1, kinda like how it works for all of the other fields in the CFAM.

I did consider removing the sibling provision/pair state, entirely, but I think it is useful it 2 cases:

It's checked for redundancy:
```
            if (!input.siblingProvisioned)
            {
                reasons.push_back(SiblingNotProvisioned);
            }
```

Technically the unprovisioned passive BMC would also fail this check:
```
            if (input.siblingCannotBeActive)
            {
                reasons.push_back(SiblingCannotBeActive);
            }
```
But we get a little more info with the check.

And then also rbmctool can then display it, which makes it easy to know that it isn't provisioned.

I'm really not that beholden to it though.

redundant-bmc/src/rbmc_tool.cpp
spinler: (I added more comments about this below)

redundant-bmc/src/active_role_handler.cpp
spinler: If I were to do it, I would be in the case:  myrole == passive and sibling role == passive and sibling provisioned == false.


If MFG absolutely requires it, I will revisit.

I also aren't that happy with how the normal simics path would be:
1. simics starts
2. Not provisioned so BMC goes to passive with error log  (I have it hardcoded to true which is why this isn't happening now)
3. simics is finally able to break in and does self and peer pairing
4. now I go to active?

PR #148, opened 2026-04-30
Add config file to the RBMC manager and use it to read the sibling present GPIO

spinler:
Add support for a config file to hold the GPIOs for sibling BMC presence and sibling reset.

The config files are all checked in under the redundant-bmc/config_files/ directory, and the one to use is specified with a meson option.

Make use of the sibling present GPIOs in the isBMCPresent function.  Since not all systems have GPIOs for this, it will fallback to what it was previously using for presence if there isn't one.

Using the config file sibling-reset GPIO instead of just hardcoding it will come in a future PR.

### DIFF
--- a/redundant-bmc/config_files/README.md
+++ b/redundant-bmc/config_files/README.md
@@ -1,0 +1,58 @@
+# Redundant BMC Configuration Files
+
+This directory contains the platform-specific redundant BMC configuration files.
+
+A config file is required, and the one to use is specified with the
+`rbmc-config-file` meson option.
+
+It is then installed into
+`/usr/share/phosphor-state-manager/redundant-bmc/config.json`.
+
+## Config File Format
+
+```json
+{
+  "sibling_bmc_reset_gpio": {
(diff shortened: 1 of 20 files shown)


redundant-bmc/config_files/README.md
```diff
@@ -0,0 +1,58 @@
...
+      "sibling_bmc_present_gpio": {
+        "name": "chassis1-present",
+        "polarity": "high"
+      }
+    }
+  ]
+}
+```
+
+## Fields
```
RameshIyyar, line 38: Do you prefer to define schema to catch as part of the CI?

RameshIyyar (approved):

redundant-bmc/config_files/README.md
spinler: The structure was pretty simple so I hadn't planned on it, but I could do that.  Will do it separately.

spinler: Opened PFEBMC-5674

PR #153, opened 2026-05-20
Add PCIe MMIO support for RBMC redundancy

RameshaR45:
Create support for transferring RBMC redundancy information through
PCIe memory mapped address space.

### DIFF
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -337,3 +337,46 @@ the following to prepare to be reset and come back as the passive BMC:
 
 1. Stop background syncing.
 2. Flush any unwritten journal messages to disk.
+
+## PCIe MMIO Redundancy State
+
+If enabled, redundancy state information is exposed to the host through PCIe
+MMIO. This allows the host to read the current redundancy state (BMC role,
+redundancy enabled status, failover status, etc.) directly from PCIe
+memory-mapped I/O space. This mechanism provides early access to redundancy
+information during host boot, before higher-level communication protocols are
+initialized, and works regardless of whether the BMC is active or passive. A
+dedicated PCIe MMIO offset is reserved for redundancy-related properties.
+
+### PCIe MMIO Layout
+
+```text
+ * @brief PCIe MMIO Redundancy State Layout
+ *
+ * Single-byte layout shared between BMC and host via PCIe MMIO.
+ *
+ * ┌─────────────────────────────────────────────────────────────┐
+ * │ Bit 7   │ Bit 6      │ Bit 5    │ Bit 4-3 │ Bit 2-0 │
+ * │ allowed │ inProgress │ enabled  │ role    │ version │
+ * └─────────────────────────────────────────────────────────────┘
+ *
+ * Layout (LSB to MSB):
+ *   Bits 0-2: version (0-7, current=1)
+ *   Bits 3-4: role (Unknown=0, Active=1, Passive=2)
+ *   Bit 5:    redundancyEnabled
+ *   Bit 6:    failoverInProgress
+ *   Bit 7:    failoversAllowed
+```
+
+### Configuration
+
+The PCIe MMIO location is configurable through the JSON configuration file. See
+[`config_files/README.md`](config_files/README.md#pcie-config-object) for
+details on the `pcie_config` configuration options.
+
+### Notes
+
+- PCIe BAR/MMIO is a standard PCIe mechanism, but the redundancy state layout
+  and offset definitions are vendor-specific.
+- The configured offsets are reserved for redundancy state information and must
+  not overlap with other PCIe MMIO users.
--- a/redundant-bmc/config_files/schema/schema.json
+++ b/redundant-bmc/config_files/schema/schema.json
@@ -18,6 +18,10 @@
             },
             "minItems": 0,
             "uniqueItems": true
+        },
+        "pcie_config": {
+            "description": "PCIe storage configuration for redundancy data. Optional - if not present, PCIe storage functionality is disabled.",
+            "$ref": "#/$defs/pcie_config"
         }
     },
     "additionalProperties": false,
@@ -53,6 +57,24 @@
                 "sibling_bmc_present_gpio": {
                     "description": "GPIO configuration for detecting sibling BMC presence",
                     "$ref": "#/$defs/gpio_config"
+                }
+            },
+            "additionalProperties": false
+        },
+        "pcie_config": {
+            "type": "object",
+            "description": "PCIe storage configuration for redundancy data",
+            "required": ["device_path", "redundancy_offset"],
+            "properties": {
+                "device_path": {
+                    "type": "string",
+                    "description": "Path to the PCIe device (e.g., '/dev/bmc-device0')",
+                    "minLength": 1
+                },
+                "redundancy_offset": {
+                    "type": "string",
+                    "description": "Offset in the PCIe device for redundancy data as a hex string (e.g., '0x3F000000')",
+                    "minLength": 1
                 }
             },
             "additionalProperties": false
--- a/redundant-bmc/src/config_data.hpp
+++ b/redundant-bmc/src/config_data.hpp
@@ -50,6 +50,24 @@ struct BMCConfig
 };
 
 /**
+ * @struct PCIeConfig
+ *
+ * Configuration for PCIe storage
+ */
+struct PCIeConfig
+{
+    /**
+     * @brief The PCIe device path (e.g., "/dev/bmc-device0")
+     */
+    std::string devicePath;
+
+    /**
+     * @brief The offset in the PCIe device for redundancy data as a hex string
+     */
+    std::string redundancyOffset;
+};
+
+/**
  * @struct RedundantBMCConfig
  *
  * Top-level configuration for redundant BMC functionality
@@ -66,6 +84,13 @@ struct RedundantBMCConfig
      *        Map key is the BMC position (0-based)
      */
     std::map<size_t, BMCConfig> bmcConfigs;
+
+    /**
+     * @brief PCIe storage configuration (optional)
+     *
+     * If not present, PCIe storage functionality is disabled
+     */
+    std::optional<PCIeConfig> pcieConfig;
 };
 
 } // namespace rbmc
--- a/redundant-bmc/src/config_parser.cpp
+++ b/redundant-bmc/src/config_parser.cpp
@@ -141,6 +141,45 @@ std::map<size_t, BMCConfig> parseBMCConfigs(const nlohmann::json& jsonData)
     return configs;
 }
 
+/**
+ * @brief Parse PCIe configuration from JSON
+ *
+ * @param[in] jsonData - The root JSON object
+ *
+ * @return Optional PCIeConfig object (empty if pcie_config not present)
+ * @throws std::runtime_error if required fields are missing or invalid
+ */
+std::optional<PCIeConfig> parsePCIeConfig(const nlohmann::json& jsonData)
+{
+    auto pcieConfigIt = jsonData.find("pcie_config");
+    if (pcieConfigIt == jsonData.end())
+    {
+        // pcie_config is optional
+        return std::nullopt;
+    }
+
+    const auto& pcieJSON = *pcieConfigIt;
+
+    auto devicePathIt = pcieJSON.find("device_path");
+    if (devicePathIt == pcieJSON.end())
+    {
+        throw std::runtime_error(
+            "pcie_config missing required 'device_path' field");
+    }
+
+    auto redundancyOffsetIt = pcieJSON.find("redundancy_offset");
+    if (redundancyOffsetIt == pcieJSON.end())
+    {
+        throw std::runtime_error(
+            "pcie_config missing required 'redundancy_offset' field");
+    }
+
+    PCIeConfig config;
+    config.devicePath = devicePathIt->get<std::string>();
+    config.redundancyOffset = redundancyOffsetIt->get<std::string>();
+    return config;
+}
+
 } // anonymous namespace
 
 RedundantBMCConfig parse(const std::filesystem::path& path)
@@ -175,6 +214,7 @@ RedundantBMCConfig parse(const std::filesystem::path& path)
             parseGPIOConfig(*resetGpioIt, "sibling_bmc_reset_gpio");
 
         config.bmcConfigs = parseBMCConfigs(jsonData);
+        config.pcieConfig = parsePCIeConfig(jsonData);
 
         return config;
     }
--- a/redundant-bmc/src/manager.cpp
+++ b/redundant-bmc/src/manager.cpp
@@ -23,7 +23,8 @@ Manager::Manager(sdbusplus::async::context& ctx,
                  std::chrono::milliseconds heartbeatInterval) :
     sdbusplus::aserver::xyz::openbmc_project::control::Failover<Manager>(
         ctx, failoverPath.c_str()),
-    ctx(ctx), redundancyInterface(ctx, *this), providers(std::move(providers)),
+    ctx(ctx), providers(std::move(providers)),
+    redundancyInterface(ctx, *this, this->providers->getPCIeStorage()),
     heartbeatInterval(heartbeatInterval)
 {
     try
--- a/redundant-bmc/src/manager.hpp
+++ b/redundant-bmc/src/manager.hpp
@@ -179,17 +179,17 @@ class Manager :
     sdbusplus::async::context& ctx;
 
     /**
+     * @brief Contains the various provider helpers
+     *
+     * NOTE: Must be declared before redundancyInterface and handler so it's
+     * destroyed after them, since their destructors access providers.
+     */
+    std::unique_ptr<Providers> providers;
+
+    /**
      * @brief The Redundancy D-Bus interface
      */
     RedundancyInterface redundancyInterface;
-
-    /**
-     * @brief Contains the various provider helpers
-     *
-     * NOTE: Must be declared before handler so it's destroyed
-     * after handler, since handler's destructor accesses providers.
-     */
-    std::unique_ptr<Providers> providers;
 
     /**
      * @brief The role handler class
--- a/redundant-bmc/src/meson.build
+++ b/redundant-bmc/src/meson.build
@@ -22,6 +22,7 @@ rbmc_lib = static_library(
     'gpio.cpp',
     'manager.cpp',
     'passive_role_handler.cpp',
+    'pcie_storage.cpp',
     'persistent_data.cpp',
     'redundancy.cpp',
     'redundancy_interface.cpp',
--- a/redundant-bmc/src/pcie_storage.cpp
+++ b/redundant-bmc/src/pcie_storage.cpp
@@ -1,0 +1,128 @@
+// SPDX-License-Identifier: Apache-2.0
+#include "pcie_storage.hpp"
+
+#include "phosphor-logging/lg2.hpp"
+
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+#include <cstring>
+#include <stdexcept>
+
+namespace pcie_data
+{
+PCIeStorageImpl::PCIeStorageImpl(const std::string& devPath, size_t offset) :
+    devicePath(devPath), redundancyOffset(offset)
+{
+    try
+    {
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg)
+        fd = open(devicePath.c_str(), O_RDWR | O_CLOEXEC);
+        if (fd == -1)
+        {
+            lg2::error("PCIe device not available at {PATH}", "PATH",
+                       devicePath);
+            return;
+        }
+
+        mmioBase = mmap(nullptr, redundancySize, PROT_READ | PROT_WRITE,
+                        MAP_SHARED, fd, static_cast<off_t>(redundancyOffset));
+        if (mmioBase == MAP_FAILED)
+        {
+            close(fd);
+            fd = -1;
+            lg2::error("PCIe mmap failed at offset {OFFSET}", "OFFSET",
+                       redundancyOffset);
+            return;
+        }
+
+        lg2::debug(
+            "PCIe storage initialized successfully at {PATH} offset {OFFSET}",
+            "PATH", devicePath, "OFFSET", redundancyOffset);
+    }
+    catch (const std::exception& e)
+    {
+        lg2::error("PCIe storage initialization failed: {ERROR}", "ERROR", e);
+    }
+}
+
+PCIeStorageImpl::~PCIeStorageImpl()
+{
+    if (mmioBase != nullptr && mmioBase != MAP_FAILED)
+    {
+        munmap(mmioBase, redundancySize);
+    }
+    if (fd != -1)
+    {
+        close(fd);
+    }
+}
+
+void PCIeStorageImpl::validateMMIO() const
+{
+    if (mmioBase == nullptr || mmioBase == MAP_FAILED)
+    {
+        throw std::runtime_error("PCIe storage not initialized");
+    }
+}
+
+void PCIeStorageImpl::writeState(const RedundancyState& state)
+{
+    std::lock_guard<std::mutex> lock(stateMutex);
+    validateMMIO();
+
+    RedundancyState modifiedState = state;
+    modifiedState.version = redundancyDataVersion;
+
+    std::memcpy(mmioBase, &modifiedState, sizeof(RedundancyState));
+    __sync_synchronize();
+}
+
+RedundancyState PCIeStorageImpl::readState()
+{
+    std::lock_guard<std::mutex> lock(stateMutex);
+    validateMMIO();
+
+    RedundancyState state{};
+    std::memcpy(&state, mmioBase, sizeof(RedundancyState));
+
+    return state;
+}
+
+void PCIeStorageImpl::updateRole(uint8_t role)
+{
+    auto state = readState();
+    state.role = role;
+    writeState(state);
+    lg2::debug("Role updated in PCIe memory to {VALUE}", "VALUE", role);
+}
+
+void PCIeStorageImpl::updateRedundancyEnabled(bool enabled)
+{
+    auto state = readState();
+    state.redundancyEnabled = enabled ? 1 : 0;
+    writeState(state);
+    lg2::debug("RedundancyEnabled updated in PCIe memory to {VALUE}", "VALUE",
+               enabled);
+}
+
+void PCIeStorageImpl::updateFailoverInProgress(bool inProgress)
+{
+    auto state = readState();
+    state.failoverInProgress = inProgress ? 1 : 0;
+    writeState(state);
+    lg2::debug("FailoverInProgress updated in PCIe memory to {VALUE}", "VALUE",
+               inProgress);
+}
+
+void PCIeStorageImpl::updateFailoversAllowed(bool allowed)
+{
+    auto state = readState();
+    state.failoversAllowed = allowed ? 1 : 0;
+    writeState(state);
+    lg2::debug("FailoversAllowed updated in PCIe memory to {VALUE}", "VALUE",
+               allowed);
+}
+
+} // namespace pcie_data
--- a/redundant-bmc/src/pcie_storage.hpp
+++ b/redundant-bmc/src/pcie_storage.hpp
@@ -1,0 +1,145 @@
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <mutex>
+#include <string>
+
+namespace pcie_data
+{
+// Default values for PCIe configuration
+constexpr auto defaultDevicePath = "/dev/bmc-device0";
+constexpr auto defaultRedundancyOffset = 0x3F00000;
+constexpr size_t redundancySize = 1;
+
+constexpr uint8_t redundancyDataVersion = 1;
+
+/**
+ * @brief PCIe MMIO Redundancy State Layout
+ *
+ * Single-byte layout shared between BMC and host via PCIe MMIO.
+ *
+ * ┌─────────────────────────────────────────────────────────────┐
+ * │ Bit 7   │ Bit 6      │ Bit 5    │ Bit 4-3 │ Bit 2-0 │
+ * │ allowed │ inProgress │ enabled  │ role    │ version │
+ * └─────────────────────────────────────────────────────────────┘
+ *
+ * Layout (LSB to MSB):
+ *   Bits 0-2: version (0-7, current=1)
+ *   Bits 3-4: role (Unknown=0, Active=1, Passive=2)
+ *   Bit 5:    redundancyEnabled
+ *   Bit 6:    failoverInProgress
+ *   Bit 7:    failoversAllowed
+ *
+ * BMC Perspective:
+ *   C bitfields in LSB-to-MSB order. BMC writes struct
+ *   directly to PCIe MMIO with first field in lowest bits.
+ *
+ * Endianness Interoperability:
+ *   Single byte eliminates endianness concerns. Both little-endian
+ *   and big-endian hosts read identical bit pattern. No byte-order conversion
+ *   needed since there's only one byte.
+ *
+ * Version field allows 7 future layout revisions.
+ */
+struct RedundancyState
+{
+    uint8_t version:3; // Bits 0-2: Layout version (0-7, current=1)
+    uint8_t role:2;    // Bits 3-4: Role (Unknown=0, Active=1, Passive=2)
+    uint8_t redundancyEnabled:1;  // Bit 5: Redundancy enabled flag
+    uint8_t failoverInProgress:1; // Bit 6: Failover in progress flag
+    uint8_t failoversAllowed:1;   // Bit 7: Failovers allowed flag
+} __attribute__((packed));
+
+static_assert(sizeof(RedundancyState) == redundancySize);
+
+/**
+ * @class PCIeStorage
+ *
+ * Abstract interface for PCIe storage operations.
+ * Allows for mocking in unit tests.
+ */
+class PCIeStorage
+{
+  public:
+    PCIeStorage() = default;
+    virtual ~PCIeStorage() = default;
+    PCIeStorage(const PCIeStorage&) = delete;
+    PCIeStorage& operator=(const PCIeStorage&) = delete;
+    PCIeStorage(PCIeStorage&&) = delete;
+    PCIeStorage& operator=(PCIeStorage&&) = delete;
+
+    /**
+     * @brief Write the complete redundancy state to PCIe storage
+     *
+     * @param[in] state - The redundancy state to write
+     */
+    virtual void writeState(const RedundancyState& state) = 0;
+
+    /**
+     * @brief Read the complete redundancy state from PCIe storage
+     *
+     * @return The redundancy state
+     */
+    virtual RedundancyState readState() = 0;
+
+    /**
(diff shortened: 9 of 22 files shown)


itsManjeet (approved):

RameshaR45:
add to approvelist

spinler:
I have a PR up that will add unit tests for that Manager class that holds this interface, so this code would need to work in unit test which I don't think it would now?

https://github.com/ibm-openbmc/phosphor-state-manager/pull/143

Let me know if you'd like some ideas.  Off the top off my head, I'd say add it to the Providers class, which there will be a mock one of in test.

redundant-bmc/src/redundancy_interface.hpp
```diff
@@ -94,6 +99,11 @@ class RedundancyInterface :
      * @brief Reference to the Manager class
      */
     Manager& manager;
+
+    /**
+     * @brief Reference to the Providers class
+     */
+    Providers& providers;
```
spinler: The class shouldn't need access to all providers.  How about just pass in PCIeStorage& ?

redundant-bmc/src/pcie_storage.cpp
spinler, line 1: I think 'pcie_storage' is probably a more appropriate name for this file so it matches the class.

redundant-bmc/src/pcie_persistent_data.cpp
```diff
@@ -0,0 +1,123 @@
...
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+#include <cstring>
+
+namespace pcie_data
+{
+
+    PCIeStorage &PCIeStorage::getInstance()
```
spinler: Ideally this code that makes this class a singleton is never merged, as it's not correct.  Can you somehow pull the proper parts of the future commit that fixes it into here?

redundant-bmc/src/pcie_persistent_data.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+    PCIeStorage::PCIeStorage()
+    {
+        try
+        {
+            
+            fd = open(DEVICE_PATH, O_RDWR | O_CLOEXEC); // Open PCIe device
+            if (fd == -1)
+            {
+                lg2::info("PCIe device not available, using filesystem only");
```
spinler: what does 'using filesystem only' mean?

redundant-bmc/src/pcie_persistent_data.cpp
```diff
@@ -0,0 +1,123 @@
...
+            }
+
+            // Memory map the PCIe region
+            mmio_base = mmap(nullptr, REDUNDANCY_SIZE, PROT_READ | PROT_WRITE,
+                             MAP_SHARED, fd, REDUNDANCY_OFFSET);
+            if (mmio_base == MAP_FAILED)
+            {
+                close(fd);
+                fd = -1;
+                lg2::info("PCIe mmap failed, using filesystem only");
```
spinler: at this point, should be an lg2::error probably.

redundant-bmc/src/pcie_persistent_data.cpp
```diff
@@ -0,0 +1,123 @@
...
+                return;
+            }
+
+            lg2::info("PCIe storage initialized successfully");
```
spinler: lg2::debug

redundant-bmc/src/pcie_persistent_data.cpp
```diff
@@ -0,0 +1,123 @@
...
+        return state;
+    }
+
+    void PCIeStorage::updateRole(uint8_t role)
+    {
+
+        auto state = readState();
+        state.role = role;
+        writeState(state);
+        lg2::info("Role updated in PCIe memory to {VALUE}", "VALUE", role);
```
spinler: debug

redundant-bmc/src/pcie_persistent_data.cpp
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    void PCIeStorage::updateDisableRedundancy(bool disable)
+    {
+
+        auto state = readState();
+        state.disable_redundancy = disable ? 1 : 0;
+        writeState(state);
+        lg2::info("DisableRedundancy updated in PCIe memory to {VALUE}", "VALUE",
```
spinler: debug

redundant-bmc/src/pcie_persistent_data.cpp
```diff
@@ -0,0 +1,123 @@
...
+                  disable);
+    }
+
+    void PCIeStorage::updateFailoverInProgress(bool inProgress)
+    {
+
+        auto state = readState();
+        state.failover_in_progress = inProgress ? 1 : 0;
+        writeState(state);
+        lg2::info("FailoverInProgress updated in PCIe memory to {VALUE}", "VALUE",
```
spinler: debug

redundant-bmc/src/pcie_persistent_data.cpp
```diff
@@ -0,0 +1,123 @@
...
+                  inProgress);
+    }
+
+    void PCIeStorage::updateFailoversAllowed(bool allowed)
+    {
+
+        auto state = readState();
+        state.failovers_allowed = allowed ? 1 : 0;
+        writeState(state);
+        lg2::info("FailoversAllowed updated in PCIe memory to {VALUE}", "VALUE",
```
spinler: debug

redundant-bmc/src/pcie_persistent_data.hpp
```diff
@@ -0,0 +1,54 @@
...
+    constexpr auto REDUNDANCY_OFFSET = 0x3F00000;
+    constexpr size_t REDUNDANCY_SIZE =
+        1;
+
+    constexpr uint8_t REDUNDANCY_DATA_VERSION = 1;
+
+    // Fixed memory layout - 8 bits total (4 bits state + 4 bits version)
+    struct RedundancyState
+    {
+        uint8_t role : 1;                 // bmc role -> Active=1, Passive=0 (1 bit)
```
spinler: the Role enum is:

    enum class Role
    {
        Unknown,
        Active,
        Passive,
    };

redundant-bmc/src/manager.hpp
```diff
@@ -163,6 +163,11 @@ class Manager :
      */
     sdbusplus::async::context& ctx;
 
+    /**
```
spinler, line 181: can you rebase your changes on latest 1210?  i had a commit that fixed this.

redundant-bmc/src/pcie_storage.cpp
```diff
@@ -7,117 +8,123 @@
...
 {
-
-    PCIeStorage &PCIeStorage::getInstance()
-    {
-        static PCIeStorage instance;
-        return instance;
-    }
-
-    PCIeStorage::PCIeStorage()
+PCIeStorageImpl::PCIeStorageImpl()
```
spinler: As I mentioned in the other commit, I would prefer this just be done initially without a need to refactor a commit that isn't merged yet.

redundant-bmc/src/meson.build
```diff
@@ -45,6 +46,8 @@ executable(
 
 executable(
     'rbmctool',
+    'pcie_storage.cpp',
```
spinler: These 2 files are already in rbmc_lib so I don't think you need toa dd them here?

redundant-bmc/src/pcie_storage.hpp
```diff
@@ -0,0 +1,78 @@
...
+    virtual ~PCIeStorage() = default;
+    PCIeStorage(const PCIeStorage&) = delete;
+    PCIeStorage& operator=(const PCIeStorage&) = delete;
+    PCIeStorage(PCIeStorage&&) = delete;
+    PCIeStorage& operator=(PCIeStorage&&) = delete;
+
+    virtual void writeState(const RedundancyState& state) = 0;
+    virtual RedundancyState readState() = 0;
+
+    virtual void updateRole(uint8_t role) = 0;
```
spinler, line 92: can you add doxygen comments like the rest of the code uses?

redundant-bmc/src/pcie_storage.hpp
```diff
@@ -0,0 +1,78 @@
...
+constexpr auto redundancyOffset = 0x3F00000;
+constexpr size_t redundancySize = 1;
+
+constexpr uint8_t redundancyDataVersion = 1;
+
+// Fixed memory layout - 8 bits total (5 bits state + 3 bits version)
+struct RedundancyState
+{
+    uint8_t role:2; // Role enum: Unknown=0, Active=1, Passive=2 (2 bits)
+    uint8_t disableRedundancy:1;  // 0=disabled, 1=enabled (1 bit)
```
spinler: So there are 2 similar redundancy properties on the Redundancy interface:
1. RedundancyEnabled
2. DisableRedundancyOverride

What you want to use is 1, RedundancyEnabled.  The 2nd one is just so that someone can turn off redundancy manually if they want, i.e. override it.

redundant-bmc/src/redundancy_interface.cpp
```diff
@@ -81,6 +103,15 @@ bool RedundancyInterface::set_property(
             "DISABLE", disable, "ERROR", e);
     }
 
+    try
```
spinler: as I mentioned above, you'll want to do this on the redundancy_enabled_t property, not this one.

redundant-bmc/src/pcie_storage.cpp
```diff
@@ -0,0 +1,130 @@
...
+        throw std::runtime_error("PCIe storage not initialized");
+    }
+
+    RedundancyState state{};
+
+    std::memcpy(&state, mmioBase, sizeof(RedundancyState));
+
+    if (state.version != redundancyDataVersion)
+    {
+        lg2::warning("Version mismatch");
```
spinler: Can you print the bad version here?

geissonator:
add to approvelist

geissonator (commented):
Lets work on the commit message a bit here. Answer the question on "why" we are doing this please. The bullet points with "what" changed are not usually necessary, that's what the code is for.

redundant-bmc/src/pcie_storage.hpp
```diff
@@ -17,7 +17,8 @@ constexpr uint8_t redundancyDataVersion = 1;
 struct RedundancyState
 {
     uint8_t role:2; // Role enum: Unknown=0, Active=1, Passive=2 (2 bits)
-    uint8_t disableRedundancy:1;  // 0=disabled, 1=enabled (1 bit)
+    uint8_t disableRedundancy:1;  // 0=redundancy enabled, 1=redundancy disabled
```
spinler: Is there a reason the polarity is this way?
I think it would be easier to understand if it followed the property rather than inverted it.

redundant-bmc/src/redundancy_interface.cpp
```diff
@@ -79,6 +79,31 @@ bool RedundancyInterface::set_property([[maybe_unused]] role_t type, Role role)
...
 
+bool RedundancyInterface::set_property(
+    [[maybe_unused]] redundancy_enabled_t type, bool enabled)
+{
+    if (enabled == redundancy_enabled())
+    {
+        return false;
+    }
+
+    lg2::info("Request to change RedundancyEnabled property to {VALUE}",
```
spinler: no trace necessary.  This is only set internally and there are already traces around it elsewhere.

spinler:
The commit message

> RBMC: Add PCIe storage support for redundant BMC sync
Implement PCIe storage interface to enable data sync between
redundant BMCs using PCIe-attached storage devices. This
provides a reliable mechanism for sharing critical state
information and configuration data between active and
passive BMCs.

This interface isn't for syncing between BMCs, it's for communicating these D-Bus properties to the host over PCIe.
And it isn't a PCIe attach storage device.
Did AI write this?

> The implementation includes:
- PCIe storage abstraction layer for sync data
- Integration with redundancy manager for automatic sync
- Persistent storage of BMC role and state information
- Error handling and recovery for storage access failures

You may as well remove this paragraph.

> This enables redundant BMC configurations to maintain
consistency across failover events by persisting and
synchronizing data through shared PCIe storage, improving
system reliability and reducing recovery time during BMC
transitions.

and this one.

You can say you added the new class that will mirror the D-Bus properties into the PCIe space so the host can read it.

redundant-bmc/src/pcie_storage.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+    if (fd != -1)
+    {
+        close(fd);
+    }
+}
+
+void PCIeStorageImpl::writeState(const RedundancyState& state)
+{
+    std::lock_guard<std::mutex> lock(stateMutex);
```
spinler: I lock_guard shouldn't be necessary because the coroutines don't do things multi-threaded, but I'm OK with it I guess in case that changes in the future.

redundant-bmc/src/pcie_storage.hpp
```diff
@@ -0,0 +1,113 @@
...
+constexpr auto redundancyOffset = 0x3F00000;
+constexpr size_t redundancySize = 1;
+
+constexpr uint8_t redundancyDataVersion = 1;
+
+// Fixed memory layout - 8 bits total (5 bits state + 3 bits version)
+struct RedundancyState
+{
+    uint8_t role:2; // Role enum: Unknown=0, Active=1, Passive=2 (2 bits)
+    uint8_t disableRedundancy:1;  // 0=redundancy disabled, 1=redundancy enabled
```
spinler: Now the comment is correct, redundancy enabled = 1, so now you can change this member name to 'redundancyEnabled', and make sure the polarity is correct below.

redundant-bmc/src/pcie_storage.hpp
```diff
@@ -0,0 +1,113 @@
...
+     * @param[in] role - The role value to set
+     */
+    virtual void updateRole(uint8_t role) = 0;
+
+    /**
+     * @brief Update only the disable redundancy field in PCIe storage
+     *
+     * @param[in] disable - The disable redundancy value to set
+     */
+    virtual void updateDisableRedundancy(bool disable) = 0;
```
spinler: and change this to updateRedundancyEnabled

spinler:
Thanks!  Looks good to me.  Would still like others to review.

I added a new method just now so you'll need to manually fix some merge conflicts that git couldn't figure out for some reason.

redundant-bmc/src/redundancy_interface.cpp
```diff
@@ -57,6 +58,49 @@ RedundancyInterface::RedundancyInterface(sdbusplus::async::context& ctx,
...
+    }
+
+    // Update PCIe memory
+    try
+    {
+        pcieStorage.updateRole(static_cast<uint8_t>(role));
+    }
+    catch (const std::exception& e)
+    {
+        lg2::info("Could not write Role to PCIe memory: {ERROR}", "ERROR", e);
```
RameshIyyar: warning trace - the same comment is applicable to all properties.

redundant-bmc/src/manager.cpp
```diff
@@ -22,7 +22,8 @@ Manager::Manager(sdbusplus::async::context& ctx,
                  std::chrono::milliseconds heartbeatInterval) :
     sdbusplus::aserver::xyz::openbmc_project::control::Failover<Manager>(
```
RameshIyyar, line 24: Commit message:

Could you explain why we are using PCIe BAR space for this instead of PLDM? Is the reason that these properties need to be accessible by the passive BMC?

Also, please add details on how the PCIe space is defined and utilized. Are we using the PCIe space exclusively for RBMC-related properties? It would be helpful to document, in the README, which properties are stored in the PCIe space and how they are mapped.

Additionally, are there any platform or implementation dependencies? How does this approach work on non-IBM systems? Should this be a configurable option?

Please also clarify how the PCIe space is standardized. If the layout is vendor-specific or implemented differently by each vendor, please document those details in the README, including any assumptions, configuration requirements.

redundant-bmc/src/pcie_storage.cpp
```diff
@@ -0,0 +1,133 @@
...
+{
+PCIeStorageImpl::PCIeStorageImpl()
+{
+    try
+    {
+        // NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg)
+        fd = open(devicePath, O_RDWR | O_CLOEXEC);
+        if (fd == -1)
+        {
+            lg2::info("PCIe device not available");
```
RameshIyyar: Error trace.

redundant-bmc/src/pcie_storage.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    RedundancyState modifiedState = state;
+    modifiedState.version = redundancyDataVersion;
+
+    uint8_t data;
+    std::memcpy(&data, &modifiedState, sizeof(RedundancyState));
+
+    std::memcpy(mmioBase, &data, sizeof(RedundancyState));
+    __sync_synchronize();
```
RameshIyyar, line 79: Why is it required?

redundant-bmc/src/pcie_storage.cpp
```diff
@@ -0,0 +1,133 @@
...
+    if (fd != -1)
+    {
+        close(fd);
+    }
+}
+
+void PCIeStorageImpl::writeState(const RedundancyState& state)
+{
+    std::lock_guard<std::mutex> lock(stateMutex);
+    if (mmioBase == nullptr || mmioBase == MAP_FAILED)
```
RameshIyyar, line 64: Why cannot you define some helper function and use it? I think you need to use this all read/write logic, right?

redundant-bmc/src/pcie_storage.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+namespace pcie_data
+{
+constexpr auto devicePath = "/dev/bmc-device0";
+constexpr auto redundancyOffset = 0x3F00000;
+constexpr size_t redundancySize = 1;
+
+constexpr uint8_t redundancyDataVersion = 1;
+
+// Fixed memory layout - 8 bits total (5 bits state + 3 bits version)
```
RameshIyyar: The version field is currently 3 bits, which means it can represent a maximum of 1 versions since you started with 1. This effectively limits the PCIe MMIO layout to 7 revisions. I assume that should be sufficient, but please confirm and document the rationale.

Additionally, should the version field be placed first in the layout so that consumers can determine the format before interpreting the remaining fields?

Could you clarify the bit ordering and endianness considerations?

Specifically:
Is the version stored in the LSBs or MSBs?
How is the layout defined from the BMC's perspective?
If the host uses a different endianness, how is interoperability ensured?

redundant-bmc/src/pcie_storage.hpp
```diff
@@ -0,0 +1,113 @@
...
+struct RedundancyState
```
RameshIyyar, line 46: Please add a option in the rbmctool to read these value from PCIe space via the CLI.

redundant-bmc/src/manager.cpp
spinler: > Should this be a configurable option?

I was on the fence if it should go in the config file if it would just be set to the same in all config files since IBM is the only user so far. But ideally it should be there, along with the address.

manojkiraneda: > Could you explain why we are using PCIe BAR space for this instead of PLDM?

These redundant properties are required very early in the PHYP boot sequence. At that stage, the PLDM stack inside PHYP is not yet initialized, so it is not possible to access this information over PLDM. This is why PCIe BAR MMIO is used as the communication mechanism instead. And for simplification of design the plan was to not run pldm stack on passive (but phyp needs this information from passive BMC as well).

> Is the reason that these properties need to be accessible by the passive BMC?

No, these properties are not specifically for the passive BMC. They are exposed in MMIO regions within the for both the active and passive BMCs.

> please add details on how the PCIe space is defined and utilized. Are we using the PCIe space exclusively for RBMC-related properties?

PCIe BAR space is used for multiple purposes, not just RBMC-related properties. It is also used for things like DMA operations, lid transfers, and other system communications (like KCS).

"Each of these uses" is mapped to specific, predefined offsets within the PCIe BAR region. In this commit  a dedicated offset range is reserved exclusively for RBMC-related properties, and nothing else uses that range.

> It would be helpful to document, in the README, which properties are stored in the PCIe space and how they are mapped.

Yes, we should update the README to clearly document which properties are stored in PCIe BAR space, along with their corresponding offset mappings and layout details.

@RameshaR45 - please add a README update describing this functionality.

> Additionally, are there any platform or implementation dependencies? How does this approach work on non-IBM systems?

Yes, this is IBM- and platform-specific. This is because:

1. The selected offsets and property definitions may overlap with or differ from what non-IBM systems use.
2. This mechanism depends on PCIe-based MMIO, while non-IBM platforms may use different interfaces for similar communication (if needed).

> Should this be a configurable option?

Yes, this should be made configurable to allow flexibility across different setups. @RameshaR45 - let’s make this configurable.

> Please also clarify how the PCIe space is standardized. If the layout is vendor-specific or implemented differently by each vendor.

PCIe BAR space itself is standardized at the PCIe specification level. However, our usage of that space in this context is not standardized.

At present, the PCIe driver is downstream-only, so full standard compliance cannot be assumed. Also, the layout we use (offsets and bit definitions) is vendor- and implementation-specific.

We will document this clearly in the README to avoid confusion.

RameshIyyar: Thanks @manojkiraneda for detailed info...Yes lets update document with details so that it would help us.

@RameshaR45 You could create a new section in the README and add relevant details.

redundant-bmc/config_files/schema/schema.json
```diff
@@ -56,6 +60,24 @@
...
+            "description": "PCIe storage configuration for redundancy data",
+            "required": ["device_path", "redundancy_offset"],
+            "properties": {
+                "device_path": {
+                    "type": "string",
+                    "description": "Path to the PCIe device (e.g., '/dev/bmc-device0')",
+                    "minLength": 1
+                },
+                "redundancy_offset": {
+                    "type": "integer",
```
spinler: We can see what others say, but maybe we go with a string here so it can be something readable, i.e. "0x3F000000".

redundant-bmc/config_files/schema/schema.json
```diff
@@ -56,6 +60,24 @@
                 }
             },
             "additionalProperties": false
+        },
+        "pcie_config": {
```
spinler, line 64: Could you also add the appropriate details to the README in config_files?

redundant-bmc/config_files/schema/schema.json
```diff
@@ -4,7 +4,7 @@
     "title": "Schema for Redundant BMC Configuration Files",
     "description": "Configuration schema for OpenBMC redundant BMC functionality",
     "type": "object",
-    "required": ["sibling_bmc_reset_gpio"],
+    "required": ["sibling_bmc_reset_gpio", "pcie_config"],
```
spinler: If some other company were to pick this up, they probably wouldn't have a hypervisor looking in the PCIe space for these, so we'd want to be able to shut it off entirely, which means this would have to be optional and if not present then the code doesn't do anything.

redundant-bmc/src/config_data.hpp
```diff
@@ -66,6 +84,11 @@ struct RedundantBMCConfig
      *        Map key is the BMC position (0-based)
      */
     std::map<size_t, BMCConfig> bmcConfigs;
+
+    /**
+     * @brief PCIe storage configuration
+     */
+    PCIeConfig pcieConfig;
```
spinler: as mentioned above, this would need to be wrapped in an std::optional.

redundant-bmc/src/providers.hpp
```diff
@@ -44,6 +46,18 @@ class Providers
...
+     * @brief Returns the PCIeStorage provider
+     */
+    virtual pcie_data::PCIeStorage& getPCIeStorage() = 0;
+
+    /**
+     * @brief Returns the config file data structures.
+     *
+     * @return The configuration
+     */
+    virtual const RedundantBMCConfig& getConfig() const = 0;
```
spinler: Missed my " RBMC: Remove Providers::getConfig" commit c9f5e06?

redundant-bmc/src/rbmc_tool.cpp
```diff
@@ -470,6 +471,61 @@ sdbusplus::async::task<> startFailover(sdbusplus::async::context& ctx,
...
+            pcie_data::defaultDevicePath, pcie_data::defaultRedundancyOffset);
+        auto state = pcieStorage.readState();
+
+        std::println();
+        std::println("PCIe MMIO Redundancy State");
+        std::println("-----------------------------");
+
+        // Decode role
+        std::string roleStr;
+        switch (state.role)
```
spinler: Could you use getPDIEnumString() here?

redundant-bmc/src/rbmc_tool.cpp
```diff
@@ -470,6 +471,61 @@ sdbusplus::async::task<> startFailover(sdbusplus::async::context& ctx,
     }
 }
 
+void displayPCIeState()
+{
+    try
+    {
+        pcie_data::PCIeStorageImpl pcieStorage(
```
spinler: Could you do something like I used above?
```
 auto config = rbmc::config_parser::readConfig();
    rbmc::SiblingResetImpl reset{ctx, config};
```

redundant-bmc/test/config_parser_test.cpp
```diff
@@ -108,7 +116,11 @@ TEST_F(ConfigParserTest, ParseEmptyBMCConfigsArray)
             "name": "sibling-bmc-reset-n",
             "polarity": "low"
         },
-        "bmc_configs": []
+        "bmc_configs": [],
+        "pcie_config": {
```
spinler, line 120: Should probably add a new testcase to parse this field.

redundant-bmc/src/pcie_storage.cpp
```diff
@@ -0,0 +1,136 @@
...
+    std::lock_guard<std::mutex> lock(stateMutex);
+    validateMMIO();
+
+    RedundancyState state{};
+    std::memcpy(&state, mmioBase, sizeof(RedundancyState));
+
+    if (state.version != redundancyDataVersion)
+    {
+        uint8_t actualVersion = state.version;
+        lg2::warning("Version mismatch, expected: {EXPECTED}, actual: {ACTUAL}",
```
RameshIyyar: Why are we proceeding further if the version is mismatch?

redundant-bmc/src/manager.cpp
```diff
@@ -22,7 +22,8 @@ Manager::Manager(sdbusplus::async::context& ctx,
...
         ctx, failoverPath.c_str()),
-    ctx(ctx), redundancyInterface(ctx, *this), providers(std::move(providers)),
+    ctx(ctx), providers(std::move(providers)),
+    redundancyInterface(ctx, *this, this->providers->getPCIeStorage()),
```
RameshIyyar, line 27: I think it will break here, please check without pcie_config from the configuration file...

redundant-bmc/src/redundancy_interface.cpp
```diff
@@ -54,9 +55,68 @@ RedundancyInterface::RedundancyInterface(sdbusplus::async::context& ctx,
         lg2::error("Failed removing NoRedundancyReasons: {ERROR}", "ERROR", e);
     }
 
+    // Sync initial D-Bus state to PCIe storage to ensure consistency
+    try
+    {
+        pcieStorage.updateFailoverInProgress(failover_in_progress_);
```
RameshIyyar: I think it will break here, PCIe config is optional so you need to make sure whether its initialized first before using it...

The same applies to all properties set logic..


Please test without PCIe config - You should just see info trace that you added and the application should work without crashing.

redundant-bmc/src/config_parser.cpp
```diff
@@ -141,6 +141,45 @@ std::map<size_t, BMCConfig> parseBMCConfigs(const nlohmann::json& jsonData)
...
+    }
+
+    auto redundancyOffsetIt = pcieJSON.find("redundancy_offset");
+    if (redundancyOffsetIt == pcieJSON.end())
+    {
+        throw std::runtime_error(
+            "pcie_config missing required 'redundancy_offset' field");
+    }
+
+    PCIeConfig config;
```
RameshIyyar, line 177: nit: You could have constructor and initialize it.

redundant-bmc/src/pcie_storage.cpp
```diff
@@ -0,0 +1,131 @@
...
+                        MAP_SHARED, fd, static_cast<off_t>(redundancyOffset));
+        if (mmioBase == MAP_FAILED)
+        {
+            close(fd);
+            fd = -1;
+            lg2::error("PCIe mmap failed at offset {OFFSET}", "OFFSET",
+                       redundancyOffset);
+            return;
+        }
+
```
RameshIyyar, line 39: I believe you should check version and add error trace (error log?) if the version is mismatch.

It could be added later also while increasing the version so I will leave it you.

RameshIyyar (approved):

manojkiraneda (approved):

spinler:
The commit message:
```
RBMC: Communicate redundancy D-Bus properties to host over PCIe
```
Isn't this over 50 chars?  That's the limit for the title, body is then 72.

redundant-bmc/src/providers_impl.hpp
```diff
@@ -71,7 +74,39 @@ class ProvidersImpl : public Providers
...
+    /**
+     * @brief Create PCIeStorage if config is present
+     *
+     * @return Optional PCIeStorageImpl
+     */
+    std::optional<pcie_data::PCIeStorageImpl> createPCIeStorage()
+    {
+        if (!config.pcieConfig.has_value())
+        {
+            lg2::info("PCIe storage not configured - pcie_config not present");
```
spinler: How about a debug trace.  If someone isn't using this feature, they won't want to see a trace about it.

redundant-bmc/src/redundancy_interface.cpp
```diff
@@ -54,9 +55,81 @@ RedundancyInterface::RedundancyInterface(sdbusplus::async::context& ctx,
...
 
+    // Sync initial D-Bus state to PCIe storage to ensure consistency
+    if (pcieStorage != nullptr)
+    {
+        try
+        {
+            pcieStorage->updateFailoverInProgress(failover_in_progress_);
+            pcieStorage->updateRedundancyEnabled(redundancy_enabled_);
+            pcieStorage->updateFailoversAllowed(failovers_allowed_);
+            // Role will be set by the role determination logic
```
spinler: I'm not sure if I missed this before or this is new, but I think we should also update the role here.  The default value is 'Unknown', and with that, PHYP will know that the code isn't ready yet, as opposed to maybe it still saying 'active' from before.

redundant-bmc/README.md
```diff
@@ -336,3 +336,59 @@ the following to prepare to be reset and come back as the passive BMC:
...
+ *
+ * Layout (LSB to MSB):
+ *   Bits 0-2: version (0-7, current=1)
+ *   Bits 3-4: role (Unknown=0, Active=1, Passive=2)
+ *   Bit 5:    redundancyEnabled
+ *   Bit 6:    failoverInProgress
+ *   Bit 7:    failoversAllowed
+```
+
+### Configuration
```
spinler, line 371: Since this is already documented in the README.md in the config_files dir, can you just link that file and point there?  That way if it changes in the future we won't have to update it in 2 places.

redundant-bmc/README.md
```diff
@@ -336,3 +336,59 @@ the following to prepare to be reset and come back as the passive BMC:
 
 1. Stop background syncing.
 2. Flush any unwritten journal messages to disk.
+
+## PCIe MMIO Redundancy State
+
+Redundancy state information is exposed through PCIe MMIO instead of PLDM
```
spinler: Maybe modify to something like:  "If enabled, redundancy state is ..."
Also, instead of just going right into it, maybe have a summary of what this is used for, which is letting the host know the redundancy state.  I wouldn't necessarily mention 'hypervisor' as that is IBM specific.

redundant-bmc/README.md
```diff
@@ -336,3 +336,59 @@ the following to prepare to be reset and come back as the passive BMC:
...
+because these properties are required very early during the Hypervisor firmware
+boot sequence, before the PLDM stack is initialized. The same information must
+also be accessible regardless of whether the BMC is active or passive.
+
+A dedicated PCIe MMIO offset is reserved for redundancy-related properties.
+While PCIe BAR space is used for multiple platform functions, this specific
```
spinler: I feel that this sentence on line 348 just rephrases line 347, so you could remove it.  I'm trying to keep the design doc fairly short since it contains a lot.  In fact, then you could move 347 up into the new summary above.

.secrets.baseline
```diff
@@ -3,7 +3,7 @@
     "files": "^.secrets.baseline$",
     "lines": null
   },
-  "generated_at": "2024-04-24T20:08:02Z",
+  "generated_at": "2026-06-11T06:42:07Z",
```
spinler: Was it necessary to check this change in?

redundant-bmc/src/providers_impl.hpp
```diff
@@ -10,6 +10,9 @@
 
 #include <phosphor-logging/lg2.hpp>
 
+#include <charconv>
```
spinler: my LSP shows charconv isn't used. Did you add it for a reason?

redundant-bmc/test/mocks/mock_providers.hpp
```diff
@@ -5,11 +5,31 @@
...
+
 namespace rbmc
 {
 
+/**
+ * @class MockPCIeStorage
+ *
+ * Mock implementation of PCIeStorage for testing.
+ */
+class MockPCIeStorage : public pcie_data::PCIeStorage
```
spinler: sorry, another one.  can you move this into a standalone files like all the other classes?

redundant-bmc/test/mocks/mock_providers.hpp
```diff
@@ -64,11 +88,18 @@ class MockProviders : public Providers
...
+        return mockPCIeStorage;
+    }
+
   private:
     MockServices mockServices;
     MockSibling mockSibling;
     MockSyncInterface mockSyncInterface;
     MockSiblingReset mockSiblingReset;
+    MockPCIeStorage mockPCIeStorage;
+    RedundantBMCConfig mockConfig;
```
spinler: I don't think that mockConfig is necessary?

redundant-bmc/src/providers.hpp
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 #pragma once
 
+#include "config_data.hpp"
```
spinler: My LSP says config_data.hpp isn't used here.

redundant-bmc/src/redundancy_interface.cpp
```diff
@@ -54,9 +55,81 @@ RedundancyInterface::RedundancyInterface(sdbusplus::async::context& ctx,
...
 
+    // Sync initial D-Bus state to PCIe storage to ensure consistency
+    if (pcieStorage != nullptr)
+    {
+        try
+        {
+            pcieStorage->updateFailoverInProgress(failover_in_progress_);
+            pcieStorage->updateRedundancyEnabled(redundancy_enabled_);
+            pcieStorage->updateFailoversAllowed(failovers_allowed_);
+            pcieStorage->updateRole(static_cast<uint8_t>(Role::Unknown));
```
spinler: You can just use 'role_' here, like the lines above.

spinler (commented):
So close on commit message, its 51.

spinler (approved):
Looks good now, thanks for all the work.

PR #158, opened 2026-05-29
RBMC: Wait 30 seconds on startup for self-pairing

spinler:
If a BMC isn't paired and so has to be passive, it would need to rebooted after it was paired in order to become active.

To create a small window where pairing can be done before the role is chosen, add a 30 second wait before determining the role where it watches for the Paired property to change to true.

Also, start checking the real value of the paired property in the 'Sibling::getPaired' function instead of just always returning true.

The action is persistent, so after it is paired for the first time the wait won't be necessary in the future.

The wait is required for the simulation model, where the pairing isn't done until several seconds after the code was looking for it.

Tested:
- If not paired, waits the 30 seconds and goes to passive
- If paired during that wait, stops waiting and goes to active
- If already paired, doesn't wait at all

The traces show up as they should:
```
21:26:48 phosphor-rbmc-state-manager[1310]: Waiting up to 30s for self pairing
21:26:55 phosphor-rbmc-state-manager[1310]: The new Paired value is True
21:26:55 phosphor-rbmc-state-manager[1310]: Starting heartbeat
```

Change-Id: I1c6607a8b3df428ee8d61b093f4b27c003843c58

### DIFF
--- a/redundant-bmc/src/manager.cpp
+++ b/redundant-bmc/src/manager.cpp
@@ -68,6 +68,8 @@ sdbusplus::async::task<> Manager::startup()
     co_await sdbusplus::async::execution::when_all(services.init(),
                                                    sibling.init());
 
+    co_await services.waitForSelfPairing();
+
     // If we know the role must be passive, set that now,
     // before starting the heartbeat or waiting for the sibling.
     auto passiveRoleInfo = co_await determinePassiveRoleIfRequired();
--- a/redundant-bmc/src/services.hpp
+++ b/redundant-bmc/src/services.hpp
@@ -243,6 +243,13 @@ class Services
      */
     virtual void setRedundancyDetermined() = 0;
 
+    /**
+     * @brief Waits for the Paired property to become true
+     *
+     * Waits up to 30 seconds.
+     */
+    virtual sdbusplus::async::task<> waitForSelfPairing() = 0;
+
   protected:
     /**
      * @brief The functions to call when the system state changes
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -9,9 +9,10 @@ redundancy related properties.
 
 ## Startup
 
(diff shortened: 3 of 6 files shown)


redundant-bmc/src/manager.cpp
```diff
@@ -68,6 +68,8 @@ sdbusplus::async::task<> Manager::startup()
     co_await sdbusplus::async::execution::when_all(services.init(),
                                                    sibling.init());
 
+    co_await services.waitForSelfPairing();
```
RameshIyyar, line 71: I’m just trying to understand why we need to wait. Can’t we simply watch the Paired property? Or are you trying to handle the initial "not paired" case, since the property defaults to false and we wouldn’t receive a signal until its value changes?

and

> The action is persistent, so after it is paired for the first time the
wait won't be necessary in the future.

What happens during pin hole reset case? Self-Pair reset?

spinler: It's because right after this call is when the BMC chooses the role, and it will go to passive if not paired, with an error log.
If I were to just wait, I would also have to write new code to go from Passive to Active without a failover, which I'm trying to avoid.

> What happens during pin hole reset case? Self-Pair reset?

That would clear the persistent indication, so it would do that wait again.

redundant-bmc/src/services.hpp
```diff
@@ -273,6 +273,13 @@ class Services
      */
     virtual void setRedundancyDetermined() = 0;
 
+    /**
+     * @brief Waits for the Paired property to become true
+     *
+     * Waits up to 30 seconds.
+     */
+    virtual sdbusplus::async::task<> waitForSelfPairing() = 0;
```
spinler, line 251: now that the mock commit merged, I need to add this to the mock class as well.

spinler: done

RameshIyyar (approved):

PR #160, opened 2026-06-01
Add helpers for accessing the external redundancy inputs

spinler:
RBMC: Add RedundancyInput data helpers

The new external redundancy inputs that are provided to the application
need to be persisted across reboots and last until the next power off.

Add helper functions around the data::read/data::write calls for them 
to simplify the calling code since they are a bit unwieldy to use.

Tested:
New testcases pass

### DIFF
--- a/redundant-bmc/src/persistent_data.cpp
+++ b/redundant-bmc/src/persistent_data.cpp
@@ -71,8 +71,9 @@ static std::string makeTimestampString(const time_t& timestamp)
 
 } // namespace util
 
-void remove(std::string_view name, const std::filesystem::path& path)
+void remove(std::string_view name)
 {
+    auto path = dataFile();
     auto json = util::readFile(path);
     if (!json)
     {
--- a/redundant-bmc/src/util.cpp
+++ b/redundant-bmc/src/util.cpp
@@ -1,0 +1,115 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#include "util.hpp"
+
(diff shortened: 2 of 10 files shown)


redundant-bmc/src/persistent_data.cpp
```diff
@@ -71,8 +71,9 @@ static std::string makeTimestampString(const time_t& timestamp)
```
RameshIyyar, line 71: Commit message:

> RMBC: Make persistent data funcs easier to test

Typo `RBMC`

redundant-bmc/src/util.cpp
```diff
@@ -0,0 +1,115 @@
...
+    }
+    catch (const std::exception& e)
+    {
+        lg2::error("Could not read external redundancy input: {ERROR}", "ERROR",
+                   e);
+    }
+    return false;
+}
+
+void writeExternalRedundancyInput(RedundancyInput input, bool value)
```
RameshIyyar: > value

confusing a bit, how about set/reset?

redundant-bmc/src/persistent_data.cpp
spinler: Thanks, done.

redundant-bmc/src/util.cpp
spinler: good idea.

spinler: done

RameshIyyar (approved):

PR #164, opened 2026-06-10
RBMC: Handle red inputs on host state changes

spinler:
The system host/power state is already tracked by RedundancyMgr.

When it changes to off, clear the redundancy inputs, and if one was cleared run determineRedundancyAndSync to try to enable redundancy again.

Also start persisting the 'host off' value, because in initSystemState the redundancy inputs should only be cleared if there was an AC loss and the way to know that is to track the last known power state.  If the last known state was on, but the current state is off, then AC must have been lost.

They shouldn't be cleared otherwise because this code also runs during a failover and there could be existing inputs that need to be used.

Tested:
- Redundancy is determined when inputs are set
- Inputs are cleared on a power off and redundancy is re-determined
- After an AC loss, inputs are cleared

Change-Id: I43ea1b2c9be434be582e30cfe5bd245cd847b44e

### DIFF
--- a/redundant-bmc/src/redundancy_mgr.cpp
+++ b/redundant-bmc/src/redundancy_mgr.cpp
@@ -4,6 +4,7 @@
 
 #include "error_data.hpp"
 #include "errors.hpp"
+#include "util.hpp"
 
 #include <persistent_data.hpp>
 #include <phosphor-logging/lg2.hpp>
@@ -11,6 +12,9 @@
 
 namespace rbmc
 {
+
+using RedundancyInput = sdbusplus::common::xyz::openbmc_project::state::bmc::
+    Redundancy::RedundancyInput;
 
 RedundancyMgr::RedundancyMgr(sdbusplus::async::context& ctx,
                              Providers& providers, RedundancyInterface& iface) :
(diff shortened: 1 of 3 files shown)


aryakpadman (approved):

redundant-bmc/src/redundancy_mgr.cpp
```diff
@@ -255,7 +301,16 @@ void RedundancyMgr::systemStateChange(SystemState newState)
...
-    determineAndSetFailoversAllowed();
+    if (newState == SystemState::off && inputsCleared)
+    {
+        lg2::info(
+            "Re-evaluating redundancy since external inputs were cleared");
+        ctx.spawn(determineRedundancyAndSync());
+    }
+    else
+    {
+        determineAndSetFailoversAllowed();
```
RameshIyyar, line 312: Do we need to reevaluate the FO allowed?

spinler: If you're asking do I need to make a call to it after line 308, it's already done inside determineRedundancyAndSync().

If you're asking why do we call it at all, it's because if the system is now in a boot we want to not allow  failovers then.

PR #168, opened 2026-06-23
RBMC:  Add getActiveFailoverBlockedReason

spinler:
Add a new getActiveFailoverBlockedReason function so that the active BMC
can check if a failover should now be blocked, and start calling it from
the ActiveRoleHandler's getFailoverBlockedReason.  This is in
preparation for failovers being started on the active BMC.

While the failover would still be forwarded to the passive BMC and its
getBlockedReason function would run, this gives an earlier, better,
failure.

Move getFailoverOptions into a util.hpp file since multiple places now
need to use it.  Also add some testcases for it.

Tested:
- New unit tests pass
- The active BMC can now block failovers

Change-Id: Ib94023960fe561c3873036b906cc563d9ef695bc
Signed-off-by: Matt Spinler <spinler@us.ibm.com>

### DIFF
--- a/redundant-bmc/src/redundancy.cpp
+++ b/redundant-bmc/src/redundancy.cpp
@@ -150,7 +150,7 @@ std::string getFailoversNotAllowedDescription(FailoversNotAllowedReason reason)
 namespace fo_blocked
 {
 
-Reason getFailoverBlockedReason(const Input& input)
+Reason getPassiveFailoverBlockedReason(const PassiveInput& input)
 {
     if (input.siblingAlive)
     {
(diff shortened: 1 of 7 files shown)


redundant-bmc/src/redundancy.cpp
```diff
@@ -219,6 +219,35 @@ Reason getPassiveFailoverBlockedReason(const PassiveInput& input)
...
+    {
+        return Reason::redundancyNotEnabled;
+    }
+
+    if (!input.failoversAllowed)
+    {
+        if (input.forceOption)
+        {
+            // Trace it but don't block it.
+            lg2::info(
```
RameshIyyar: warning?

RameshIyyar (approved):

redundant-bmc/src/redundancy.cpp
spinler: yea, good idea

PR #169, opened 2026-06-25
RBMC: Keep track of in-progress waits

spinler:
Introduce the ability to keep track of when the code is waiting for something to happen, which happens a lot while dealing with redundant BMCs.

The active waits are then stored in a separate file so rbmctool can display them.

This is done with a new 'WaitTracker' class, and a WaitGuard class that uses RAII to have an active wait while it is in scope.

rbmctool can then show them like:

```
Local BMC
-----------------------------
Role:                 Active
BMC Position:         0
Redundancy Enabled:   false
BMC State:            NotReady
Failovers Allowed:    false
Failover In Progress: false
FW Version Hash:      31BE0048
Peer Connected:       NotDetermined
Role Reason:          Resuming previous role
    In transition
Active Waits:  <--------------------------------- 
    StartUnit
Last failover driven by this BMC:
    Requester: Tool
    Timestamp: 2026-06-24 15:59:11 UTC

```

### DIFF
--- a/redundant-bmc/src/meson.build
+++ b/redundant-bmc/src/meson.build
@@ -34,6 +34,7 @@ rbmc_lib = static_library(
     'sync_interface_impl.cpp',
     'system_state.cpp',
     'util.cpp',
+    'wait_tracker.cpp',
     dependencies: rbmc_deps,
 )
 
--- a/redundant-bmc/src/wait_tracker.cpp
+++ b/redundant-bmc/src/wait_tracker.cpp
@@ -1,0 +1,223 @@
+// SPDX-License-Identifier: Apache-2.0
+#include "wait_tracker.hpp"
+
+#include <nlohmann/json.hpp>
+#include <phosphor-logging/lg2.hpp>
+
+#include <fstream>
+
+namespace rbmc
+{
+
+constexpr auto waitStatusFileName = "wait_status.json";
+
+std::string waitOperationToString(WaitOperation op)
+{
+    std::string str{"Unknown"};
+    switch (op)
+    {
+        case WaitOperation::systemInventoryPath:
+            str = "SystemInventoryPath";
+            break;
+        case WaitOperation::systemInventoryStatus:
+            str = "SystemInventoryStatus";
+            break;
+        case WaitOperation::peerConnection:
+            str = "PeerConnection";
+            break;
+        case WaitOperation::selfPairing:
+            str = "SelfPairing";
+            break;
+        case WaitOperation::startUnit:
+            str = "StartUnit";
+            break;
+        case WaitOperation::siblingAlive:
+            str = "SiblingAlive";
+            break;
+        case WaitOperation::siblingBMCSteadyState:
+            str = "SiblingBMCSteadyState";
+            break;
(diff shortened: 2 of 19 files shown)


spinler:
this is fairly similar to my previous attempt.  Now it writes to a wait_status.json file (which can't be synced), and also has a rbmctool -w option which says how much time has elapsed for a wait and what the timeout is.

aryakpadman (approved):

redundant-bmc/src/meson.build
```diff
@@ -34,6 +34,7 @@ rbmc_lib = static_library(
     'sync_interface_impl.cpp',
     'system_state.cpp',
     'util.cpp',
+    'wait_tracker.cpp',
```
RameshIyyar, line 37: > WaitTracker::WaitGuard guard(waitTracker, WaitOperation::startUnit, timeout);

Could we also add an argument to pass metadata (similar to user data in callback-based designs)? This would allow the caller registering the wait tracker to provide additional context. We could include that information in the JSON output and display it through `rbmctool`, making it easier for users to understand what the wait is for.

I suggest adding an optional string argument for this purpose.

redundant-bmc/src/wait_tracker.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+        nlohmann::json j;
+        nlohmann::json::array_t waitsArray;
+
+        for (const auto& [id, info] : activeWaits)
+        {
+            nlohmann::json waitObj;
+            waitObj["id"] = id;
+            waitObj["operation"] = waitOperationToString(info.operation);
+            waitObj["operationEnum"] = static_cast<int>(info.operation);
```
RameshIyyar, line 148: is it really required?

spinler: hmm, I guess not.  will remove.

redundant-bmc/src/meson.build
spinler: Other than 'StartUnit', I'm not really seeing where the metadata would help much, as the enum name is basically what is printed when it shows the wait.

```
    systemInventoryPath,
    systemInventoryStatus,
    peerConnection,
    selfPairing,
    startUnit,
    siblingAlive,
    siblingBMCSteadyState,
    siblingHealthTimer,
    peerConnectionTimer,
    bmcResetTimer,
    fullSync
```

And for StartUnit there's really just one choice per BMC.

redundant-bmc/src/wait_tracker.cpp
spinler: oh, it's "operation" that isn't used.  I had just kept in in for readability but probably not really that necessary.  will remove.

redundant-bmc/src/meson.build
RameshIyyar: Hmm ok, I was thinking only about StartUnit only..we could revisit if its not worth now.

spinler: Yea, for now I'd rather keep it how it is, and can revisit in the future if necessary.  Could also I guess pass in the wait enum to startUnit and use different ones.

RameshIyyar (approved):

PR #170, opened 2026-06-29
RBMC: On passive, wait for full sync before F.O.

spinler:
Manager::doFailoverFromPassive will delete the PassiveRoleHandler object as part of its transition to being active as part of a failover.

If a full sync was in progress then, PassiveRoleHandler::startSync would still be running, and eventually when the sync completes there is a call to watchSyncHealth() which takes a 'this' pointer, causing a use after free crash.

To avoid this, do the following:

1. Don't delete the PassiveRoleHandler object until after the failover imminent delay is done.  Instead, just stop its watches on the active BMC.  With just this change that should help avoid the situation.

2. While FailoverImminent is still active, check if the full sync is still in progress and wait for it to complete before continuing. This way, the delay, which is already expected, may just be extended for a bit, and no other behavior is changed.

The wait can't happen after the other BMC is reset, because if it does somehow take a long time, it can complete its reboot and just come back and take the active role again, since this BMC is alive and passive.

Tested:

Continuous failover testing completes.

Change-Id: I5aacaab71dc3538341fd0deb78b95268fa908d7c

### DIFF
--- a/redundant-bmc/src/manager.cpp
+++ b/redundant-bmc/src/manager.cpp
@@ -452,6 +452,25 @@ sdbusplus::async::task<> Manager::method_call(start_failover_t /* unused */,
     }
 }
 
+sdbusplus::async::task<> Manager::preFailoverFullSyncCheck()
+{
+    if (!providers->getSyncInterface().isFullSyncInProgress())
+    {
+        co_return;
+    }
+
+    lg2::info(
+        "Waiting for in progress full sync to complete before continuing failover");
+
+    using namespace std::chrono_literals;
+    while (providers->getSyncInterface().isFullSyncInProgress())
+    {
+        co_await sdbusplus::async::sleep_for(ctx, 500ms);
+    }
+
+    lg2::info("Full sync completed, proceeding with failover");
+}
+
 // NOLINTNEXTLINE
 sdbusplus::async::task<> Manager::doFailoverFromPassive(Requester requester)
 {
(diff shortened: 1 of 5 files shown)


redundant-bmc/src/manager.cpp
```diff
@@ -432,6 +432,25 @@ sdbusplus::async::task<> Manager::method_call(start_failover_t /* unused */,
...
+    lg2::info(
+        "Waiting for in progress full sync to complete before continuing failover");
+
+    using namespace std::chrono_literals;
+    while (providers->getSyncInterface().isFullSyncInProgress())
+    {
+        co_await sdbusplus::async::sleep_for(ctx, 500ms);
+    }
+
+    lg2::info("Full sync completed, proceeding with failover");
```
aryakpadman, line 471: Do we need to reject the requested FO , if the ongoing full sync fails for some reason?

spinler: A few thoughts here:

The failover could have been issued because something was really wrong with the previous active, so even if syncs don't work, it's still better to have this BMC as active.

When we start failing over from the active BMC, it would be tough to reject a failover after it already started.

aryakpadman: Yeah agree.
Thanks for clarifying

aryakpadman (approved):

RameshIyyar (approved):

spinler:
Had to fix a minor merge conflict in the header before merging.

PR #171, opened 2026-07-01
 RBMC: Start failover on active BMC

spinler:
Add support for the StartFailover method call to work on the active BMC.
When called on the active, after it gets by the failover blocked checks,
it will call the StartFailover method on the Failover interface that is
on the '/xyz/openbmc_project/state/bmc1' object path hosted by the
sibling daemon.

This will forward the failover request to the passive BMC to do a normal
failover from there.

The active BMC should get reset as part of the failover, if that doesn't
happen and the BMC stays up, an error log will be created.  That's done
by starting a 30 second timer after the failover request was made.

Tested:
Failover starts successfully:

```
phosphor-rbmc-state-manager[869]: Failover requester is Tool
phosphor-rbmc-state-manager[869]: Failover being started from the active BMC
phosphor-rbmc-state-manager[869]: Sibling property FailoverImminent changed
phosphor-rbmc-state-manager[869]: A failover is imminent
phosphor-rbmc-state-manager[869]: Starting journal flush
phosphor-rbmc-state-manager[869]: Completed journal flush with rc 0
```

And then the BMC Resets.

### DIFF
--- a/redundant-bmc/src/active_role_handler.cpp
+++ b/redundant-bmc/src/active_role_handler.cpp
@@ -351,7 +351,6 @@ void ActiveRoleHandler::siblingFailoverImminent(bool imminent)
     {
         lg2::warning("A failover is imminent");
 
-        ctx.spawn(providers.getSyncInterface().disableBackgroundSync());
         ctx.spawn(providers.getServices().flushJournal());
     }
 }
--- a/redundant-bmc/README.md
+++ b/redundant-bmc/README.md
@@ -258,6 +258,11 @@ not be re-enabled afterwards, depending on if something is preventing it or not.
 The failover is always driven by the original passive BMC, which then takes over
 as active.
 
+Though driven by the passive BMC, a failover can still be initiated on the
+active BMC by calling its `StartFailover` D-Bus method. After checking for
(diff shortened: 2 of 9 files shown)


spinler:
At the time of putting this up, the Failover interface doesn't exist yet on the sibling daemon on state/bmc1.   I'll put that PR up after https://github.ibm.com/openbmc/openpower-proc-control/pull/141 goes in.

redundant-bmc/src/active_role_handler.cpp
```diff
@@ -351,7 +351,6 @@ void ActiveRoleHandler::siblingFailoverImminent(bool imminent)
     {
         lg2::warning("A failover is imminent");
 
-        ctx.spawn(providers.getSyncInterface().disableBackgroundSync());
```
RameshIyyar, line 354: Makes sense. We are not planning to trigger failover (FO) due to network issues between the BMCs. If such an issue occurs, we will disable data synchronization first, and if the peer connection retries continue to fail, we will disable redundancy so its safe to run background sync when a failover is imminent and later we will disable sync as part of the FO process and attempt to sync it back.

RameshIyyar (approved):

PR #173, opened 2026-07-15
RBMC: fix RBMC PCIe redundancy offset to avoid MCTP collision

manojkiraneda:
The redundancy data offset was set to 0x3F00000, which collides with the MCTP base offset. Move it to 0x3EFFF80 instead.

MCTPBASE (0x3F00000) - CACHELINE SIZE(0x80) = 0x3EFFF80

Update all config files, tests, schema, and documentation accordingly.

Since now the address is not page aligned, need to also page alignment magic for the mmap to succeed.

Tested by :

```
root@huygens:/tmp# rbmctool --read-pcie
<7> PCIe storage initialized successfully at /dev/bmc-device0 offset 0x3efff80

PCIe MMIO Redundancy State
-----------------------------
Version:              1
Role:                 Active
Redundancy Enabled:   false
Failover In Progress: false
Failovers Allowed:    false
Raw Byte Value:       0b00001001 (0x09)

```

### DIFF
--- a/redundant-bmc/config_files/README.md
+++ b/redundant-bmc/config_files/README.md
@@ -34,7 +34,7 @@ It is then installed into
   ],
   "pcie_config": {
     "device_path": "/dev/bmc-device0",
-    "redundancy_offset": "0x3F00000"
+    "redundancy_offset": "0x3EFFF80"
   }
 }
 ```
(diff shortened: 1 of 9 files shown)


RameshIyyar (approved):

manojkiraneda:
I am still yet to test this on simics, will post my testing updates here.

spinler (approved):

geissonator:
add to approvelist

PR #175, opened 2026-07-16
RBMC: Track code updates

spinler:
The plan is to do code updates in sequence, updating and rebooting the active BMC, following by updating and rebooting the passive BMC.  After the active BMC is rebooted into its new level, redundancy will be disabled because the code levels don't match.

Normally, an error log would be created for this.  This PR adds the functionality to know that it's in the middle of the code update process and then skip the error log.  It does this by:

1. Watch for the Activation property on any Software.Activation interface to go to 'Activating'.
2. Verify this is a BMC code update by checking the Purpose property on the Version interface on that path.
3. Persist the code update in progress indication
4. On a reboot, keep this indication
5. If set, then when determining redundancy if the only reason for no-redundancy is 'mismatched levels', then don't create an error.
6. If redundancy can be enabled, or if that reason isn't listed, clear the indication.
7. On a boot, also a clear the indication, this was a policy decision made that the customer is no longer interested in the updates and would rather boot.

In the future, the other BMC will also want to know this indication so that it can be more lenient on allowing failovers in the case of a bad code update.  To accommodate that, also put this indication on D-Bus using the Software.Activation interface on state/bmc0.  It will have two values only, Activating or Active.

### DIFF
--- a/redundant-bmc/src/redundancy_mgr.cpp
+++ b/redundant-bmc/src/redundancy_mgr.cpp
@@ -17,10 +17,16 @@ using RedundancyInput = sdbusplus::common::xyz::openbmc_project::state::bmc::
     Redundancy::RedundancyInput;
 
 RedundancyMgr::RedundancyMgr(sdbusplus::async::context& ctx,
-                             Providers& providers, RedundancyInterface& iface) :
+                             Providers& providers, RedundancyInterface& iface,
+                             CodeUpdateActivation& codeUpdateActivation) :
     ctx(ctx), providers(providers), redundancyInterface(iface),
+    codeUpdateActivation(codeUpdateActivation),
     manualDisable(iface.disable_redundancy_override())
-{}
+{
+    providers.getServices().addCodeUpdateCallback(
+        Role::Active,
+        std::bind_front(&RedundancyMgr::codeUpdateActivationChanged, this));
+}
 
 void RedundancyMgr::determineAndSetRedundancy()
 {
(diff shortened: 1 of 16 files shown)


redundant-bmc/src/services_impl.cpp
```diff
@@ -470,6 +479,83 @@ sdbusplus::async::task<> ServicesImpl::watchPairingPropertiesChanged(
...
 
+sdbusplus::async::task<bool> ServicesImpl::isBMCCodeUpdate(
+    const std::string& path) const
+{
+    try
+    {
+        auto service = co_await util::getService(ctx, path, Version::interface);
+        auto purpose =
+            co_await Version(ctx).service(service).path(path).purpose();
+        co_return purpose == Version::VersionPurpose::BMC;
```
RameshIyyar, line 490: I saw its [deprecated property](https://github.com/openbmc/phosphor-dbus-interfaces/tree/master/yaml/xyz/openbmc_project/Software), is it safe to use?

aryakpadman (approved):

redundant-bmc/src/redundancy_mgr.cpp
```diff
@@ -39,17 +45,38 @@ void RedundancyMgr::determineAndSetRedundancy()
...
+    {
+        // If a code update was in progress, but both BMCs were updated so
+        // the versions are the same again, consider it complete.
+        if (redundancyInterface.redundancy_enabled() ||
+            !std::ranges::contains(reasons,
+                                   ReasonForNoRedundancy::CodeVersionMismatch))
+        {
+            lg2::info("Clearing code-update-in-progress indication");
+            codeUpdateActivation.clearCodeUpdateInProgress();
+        }
```
aryakpadman, line 59: nit : If we add a warning log indicating that the peer BMC need to undergo a code update as else block, will it add value?

aryakpadman (approved):

redundant-bmc/src/services_impl.cpp
```diff
@@ -470,6 +479,83 @@ sdbusplus::async::task<> ServicesImpl::watchPairingPropertiesChanged(
...
+    }
+    catch (const sdbusplus::exception_t& e)
+    {
+        lg2::debug(
```
RameshIyyar: error trace?

redundant-bmc/src/redundancy_mgr.cpp
```diff
@@ -39,17 +45,38 @@ void RedundancyMgr::determineAndSetRedundancy()
...
+    }
+
     if (!redundancyInterface.redundancy_enabled())
     {
         auto wasManuallyDisabled = std::ranges::contains(
             oldReasons, ReasonForNoRedundancy::ManuallyDisabled);
 
+        auto mismatchOnly =
```
RameshIyyar: Just a suggestion: codeMismatchOnly

redundant-bmc/src/redundancy_mgr.cpp
```diff
@@ -301,6 +342,18 @@ void RedundancyMgr::systemStateChange(SystemState newState)
...
         }
     }
+    else if (newState == SystemState::booting)
+    {
+        // Consider a boot the end of allowing special treatment
+        // for code updates, as that must have been deemed more
+        // important than finishing the code update completely.
+        if (codeUpdateActivation.codeUpdateInProgress())
+        {
+            lg2::warning(
```
RameshIyyar: is it should be warning? its expected normal flow, right?

redundant-bmc/src/redundancy_mgr.cpp
spinler: I think since redundancy is re-calculated anything something changes, this may run more than once in this situation, so I'd prefer to not add extra traces here.

redundant-bmc/src/redundancy_mgr.cpp
spinler: sure, can change.

redundant-bmc/src/redundancy_mgr.cpp
spinler: I was thinking that it isn't expected that someone would boot before they updated the passive BMC.  The normal flow would be that this gets cleared after the passive BMC is rebooted into the same code level as the active.

That being said, it's not like this being a warning would actually make someone notice it.  I can change it to info.

redundant-bmc/src/services_impl.cpp
spinler: Good to know.  I'll investigate.

redundant-bmc/src/services_impl.cpp
spinler: Yea, that's probably fine.  Will change.

redundant-bmc/src/services_impl.cpp
spinler: I looked at the updater code.  The Version interface isn't going anywhere on it, and will still be there after the meson option to enable the new updater daemon is enabled.

This is easier than following an association, and then checking if that endpoint has a inventory.BMC interface on it (which is currently broken for due to service deps), so will keep this unless I am forced to move for some reason.

RameshIyyar (approved):

PR #177, opened 2026-07-17
RBMC: Increase active target timeout

spinler:
In a large simulation system, starting the active target took over 16 minutes.  To be extra safe, just increase the timeout to 30.  Also make a constant for it.

Change-Id: I05a2b67a17092bd40ca5331dda524266c4ff8071

### DIFF
--- a/redundant-bmc/src/active_role_handler.cpp
+++ b/redundant-bmc/src/active_role_handler.cpp
@@ -13,6 +13,7 @@ namespace rbmc
 using Failover = sdbusplus::common::xyz::openbmc_project::control::Failover;
 
 constexpr auto bmcActiveTarget = "obmc-bmc-active.target";
+const std::chrono::minutes bmcActiveTargetTimeout{30};
 const std::chrono::minutes siblingHealthTimeout{5};
 
 // NOLINTNEXTLINE
@@ -44,7 +45,7 @@ sdbusplus::async::task<> ActiveRoleHandler::start()
     try
     {
         // NOLINTNEXTLINE(clang-analyzer-core.uninitialized.Branch)
-        co_await services.startUnit(bmcActiveTarget, std::chrono::minutes{10});
+        co_await services.startUnit(bmcActiveTarget, bmcActiveTargetTimeout);
     }
     catch (const std::exception& e)
     {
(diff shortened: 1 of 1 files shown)


redundant-bmc/src/active_role_handler.cpp
```diff
@@ -13,6 +13,7 @@ namespace rbmc
 using Failover = sdbusplus::common::xyz::openbmc_project::control::Failover;
 
 constexpr auto bmcActiveTarget = "obmc-bmc-active.target";
+const std::chrono::minutes bmcActiveTargetTimeout{30};
```
RameshIyyar, line 16: I think the timeout should be increased because the pairing process takes longer in large-node simulation environments.

However, I believe there was a suggestion to always use a known CA certificate in the simulation model, in which case the pairing process would not be required. If we go with that approach, there would be no need to increase the timeout. That said, increasing the timeout wouldn't cause any issues either, so I'll leave to you.

RameshIyyar (approved):

redundant-bmc/src/active_role_handler.cpp
spinler: I saw this the other day:
```
Dec 19 21:27:59 huygens phosphor-rbmc-state-manager[2680]: Waiting up to 30s for self pairing
Dec 19 21:28:29 huygens phosphor-rbmc-state-manager[2680]: Timed out waiting for self pairing
Dec 19 21:28:29 huygens phosphor-rbmc-state-manager[2680]: Role = xyz.openbmc_project.State.BMC.Redundancy.Role.Passive due to: BMC is not paired
Dec 19 21:28:29 huygens phosphor-rbmc-state-manager[2680]: Starting heartbeat
Dec 19 21:28:52 huygens phosphor-rbmc-state-manager[2680]: Setting 'SiblingCannotBeActive' in ReasonsForNoRedundancy property
Dec 19 21:28:52 huygens phosphor-rbmc-state-manager[2680]: No redundancy because: SiblingCannotBeActive
Dec 19 21:29:15 huygens phosphor-rbmc-state-manager[2680]: Starting unit obmc-bmc-passive.target
Dec 19 21:29:37 huygens phosphor-rbmc-state-manager[2680]: The new Paired value is True
```

Which took a minute and 38 seconds.  But usually it was less than 40.

PR #180, opened 2026-07-29
RBMC: Add Manager testcases

spinler:
Add testcases for the Manager class where the BMC becomes passive on startup.

The mock classes all have a new setupDefaultBehaviors function to handle returning default values for their coroutine functions so the testcases themselves don't have to specify them unless they want a different one.

There is also a ProgressTracker base and child class that lives on Providers so that the code can set a progress point that the testcase can wait for, like finishing starting the active or passive role handlers.  This way, the testcase can just wait for those to be set and can be as fast as possible, as opposed to just delaying for some amount of time.  In the future, there would also be one for a failover completion.

Change-Id: I904b66138dc12c89f9afda9925192d162bd9eff3

### DIFF
--- a/redundant-bmc/test/manager_test.cpp
+++ b/redundant-bmc/test/manager_test.cpp
@@ -1,23 +1,503 @@
 // SPDX-License-Identifier: Apache-2.0
+#include "manager.hpp"
+#include "mocks/async_helpers.hpp"
 #include "mocks/mock_providers.hpp"
+#include "persistent_data_test_fixture.hpp"
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
 using namespace rbmc;
+using namespace testing;
+using namespace test_helpers;
 
-class ManagerTest
+using Redundancy =
+    sdbusplus::common::xyz::openbmc_project::state::bmc::Redundancy;
+
+constexpr std::chrono::milliseconds hbInterval{5};
+
+class ManagerTest : public rbmc::test::PersistentDataTestFixture
 {
   protected:
-    void SetUp()
+    struct TestScenarioConfig
+    {
+        bool paired = true;
+        std::optional<size_t> bmcPosition = 0;
+        bool systemInventoryAvailable = true;
+        bool siblingServiceRunning = true;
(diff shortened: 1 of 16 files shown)


redundant-bmc/test/manager_test.cpp
```diff
@@ -1,23 +1,454 @@
...
+                                   errors::Level::Error, _))
+        .Times(0);
+
+    EXPECT_CALL(services,
+                startUnit("obmc-bmc-passive.target", passiveTargetTimeout))
+        .Times(1);
+
+    RedundancyProps expectedProps{
+        .role = Role::Passive,
+        .redEnabled = false,
```
RameshIyyar, line 392: Its should be enabled since `reasonsForNoRedundancy` is empty, right?

redundant-bmc/test/manager_test.cpp
```diff
@@ -1,23 +1,454 @@
...
+    // for sibling role during startup
+    EXPECT_CALL(sibling, waitForSiblingRole()).Times(1);
+
+    EXPECT_CALL(services,
+                startUnit("obmc-bmc-passive.target", passiveTargetTimeout))
+        .Times(1);
+
+    RedundancyProps expectedProps{
+        .role = Role::Passive,
+        .redEnabled = false,
```
RameshIyyar, line 490: Its should be enabled since `reasonsForNoRedundancy` is empty, right?

spinler: in this case the other BMC would still have:
```
.siblingRole = Role::Unknown
```

so redundancy wouldn't be enabled here.

redundant-bmc/test/manager_test.cpp
spinler: there's nothing here setting redundancy to enabled, so it would still be disabled.  I'll look at a new test for that.

spinler: Added BecomesPassive_SiblingAlreadyActive_RedEnabled

RameshIyyar (approved):

PR #188, opened 2026-08-04
Phosphor-state-manager: rebase with upstream

danielhufnagle:
Rebase phosphor-state-manager with upstream. Tested by building openbmc with the rebase and then testing against a huygens simics session as well as rbmc prototype hardware.


geissonator (approved):
Reviewed all downstream IBM commits and verified they are still present (in some cases merged) within this rebase fork.

PR #189, opened 2026-08-05
RBMC: Use sdbusplus-generated property accessors

ArchanaKakani:
Replace stale trailing-underscore Redundancy property references with the generated sdbusplus accessors/setters.

These changes are required to support latest sdbusplus changes added in [1].

This keeps the existing *_t property tag types unchanged and fixes the build failure caused by references such as role_, failover_in_progress_,and disable_redundancy_override_ no longer being present.

[1]: https://github.com/openbmc/sdbusplus/commit/b1b8fac883af0a1db0bc74ac68877cbb02c3ad41

### DIFF
--- a/redundant-bmc/src/redundancy_interface.cpp
+++ b/redundant-bmc/src/redundancy_interface.cpp
@@ -21,7 +21,7 @@ RedundancyInterface::RedundancyInterface(sdbusplus::async::context& ctx,
 {
     try
     {
-        disable_redundancy_override_ =
+        properties.disable_redundancy_override =
             data::read<bool>(data::key::disableRed).value_or(false);
     }
     catch (const std::exception& e)
@@ -32,9 +32,9 @@ RedundancyInterface::RedundancyInterface(sdbusplus::async::context& ctx,
 
     try
     {
-        failover_in_progress_ =
+        properties.failover_in_progress =
             data::read<bool>(data::key::failoverInProgress).value_or(false);
-        if (failover_in_progress_)
+        if (properties.failover_in_progress)
         {
             lg2::info("Failover was previously in progress");
         }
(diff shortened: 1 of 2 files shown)


redundant-bmc/src/redundancy_interface.cpp
```diff
@@ -21,8 +21,8 @@ RedundancyInterface::RedundancyInterface(sdbusplus::async::context& ctx,
 {
     try
     {
-        disable_redundancy_override_ =
-            data::read<bool>(data::key::disableRed).value_or(false);
+        disable_redundancy_override(
```
spinler: this might emit signals.  I'd rather use the new functionality that is replacing the single members:

In the aserver/common headers:
```

    struct properties_t
    {
        Role role = Role::Unknown;
        bool redundancy_enabled = false;
        bool disable_redundancy_override = false;
        bool failovers_allowed = false;
        bool failover_imminent = false;
        bool failover_in_progress = false;
        size_t redundancy_minimum = 2;
        size_t redundancy_maximum = std::numeric_limits<size_t>::max();
        size_t functional_minimum = 1;
        std::vector<DisabledRedundancyReason> disabled_redundancy_reasons = {};
        FailoversNotAllowedReason failovers_not_allowed_reason = FailoversNotAllowedReason::None;

        /** Unpack properties from a map-like (name to PropertiesVariant). */
        template <typename M>
        static properties_t unpack(const M&);
    };
```
and
```
  protected:
    properties_t properties{};
```

redundant-bmc/src/redundancy_interface.hpp
```diff
@@ -1,10 +1,12 @@
 // SPDX-License-Identifier: Apache-2.0
 #pragma once
+#define SDBUSPLUS_ASYNC_NEW_PROPERTY_MEMBERS
```
spinler: Are you sure this is necessary?  I thought this was only temporary upstream.

redundant-bmc/src/redundancy_interface.cpp
```diff
@@ -21,7 +21,7 @@ RedundancyInterface::RedundancyInterface(sdbusplus::async::context& ctx,
 {
     try
     {
-        disable_redundancy_override_ =
+        this->properties.disable_redundancy_override =
```
spinler: Nit: This codebase doesn't use the 'this->' for class members, since it' isn't strictly necessary.  Could you remove?

spinler (approved):

PR #193, opened 2026-08-17
obmc-power-start-pre: fix repeated start of non-instanced services

geissonator:
With multi-chassis SMP, obmc-power-start-pre@.target is started once per chassis. The mapper-subtree-remove services it pulled in are non-instanced, so starting the target 8 times caused systemd to hit the rate limit on those services and fail.

Move the mapper-subtree-remove Wants/After dependencies into a new non-instanced service, phosphor-wait-poweron-blocks.service, which is wired exclusively into obmc-chassis-poweron@0.target via RequiredBy and ordered Before obmc-power-start-pre@0.target so the mapper cleanup still completes before the pre-power target runs.

Tested:
- Confirmed a power on with multiple chassis does not cause target to fail with this fix
- Confirmed if a Logging.ErrorBlocksTransition is found on dbus then poweron is not allowed

Change-Id: I8fe19af303ea91e7f06190d4021fbd2192e2af3a


geissonator:
Upstream commit: https://gerrit.openbmc.org/c/openbmc/phosphor-state-manager/+/93426

spinler (approved):

PR #195, opened 2026-08-18
multi-chassis-smp: clear chassis@0-on file on aggregate power on

geissonator:
The standard chassis_state_manager clears the CHASSIS_ON_FILE (e.g. /run/openbmc/chassis@N-on) once it detects that the corresponding obmc-chassis-poweron@N.target has completed via the JobRemoved signal. This file is a temporary indicator used by chassis-related systemd services to detect that chassis power was already on when the BMC rebooted, so those services can skip running. This file must be cleared once the corresponding systemd target has completed otherwise future power on requests will not actually power on the system.

The SMP aggregator (chassis instance 0) never cleared its own chassis@0-on file. Add a JobRemoved signal match and handler (sysStateChangeJobRemoved) that removes /run/openbmc/chassis@0-on when obmc-chassis-poweron@0.target completes with result "done".

The cleanup must happen in the JobRemoved handler rather than during the initial aggregatePowerState() call at startup. The two events race: phosphor-reset-chassis-running@0.service writes the file at the same time the aggregator is doing its initial D-Bus property scan. If we clear the file during aggregation the service finishes shortly afterwards and re-creates it, leaving it behind permanently. By waiting for the poweron target's JobRemoved signal we guarantee the file was already written before we remove it.

Tested:
- Verified /run/openbmc/chassis@0-on is removed after obmc-chassis-poweron@0.target completes following a BMC reboot with chassis power already on

Change-Id: I704c6a583871c470355445624692bcef622bf0ee

### DIFF
--- a/chassis_state_manager_smp.cpp
+++ b/chassis_state_manager_smp.cpp
@@ -10,6 +10,7 @@
 #include <xyz/openbmc_project/Common/error.hpp>
 #include <xyz/openbmc_project/Inventory/Item/common.hpp>
 
+#include <filesystem>
 #include <format>
 #include <fstream>
 #include <iostream>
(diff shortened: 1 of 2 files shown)


geissonator:
Upstream commit: https://gerrit.openbmc.org/c/openbmc/phosphor-state-manager/+/93468

spinler (approved):

gtmills:
Request from @andrewg
Pull Request: https://github.com/ibm-openbmc/phosphor-state-manager/pull/195
Upstream: Commit merged / being reviewed upstream
Upstream Information: https://gerrit.openbmc.org/c/openbmc/phosphor-state-manager/+/93468
Fixes: N/A
Releases: 1210
Release List comments: Multi chassis specific so P12 only
Release Note: Clear chassis@0-on file on aggregate power on

PR #198, opened 2026-08-24
utils: return log entry path from createError and link to BMC dump

devenrao:
Multiple repositories create user dumps in response to system errors. On a running system it is difficult to identify which component created a particular dump and what error drove it. By capturing the associated error log details inside the dump archive, a user dump can be directly correlated with the error that triggered it.

The Logging.Create.Create() D-Bus method already returns the object path of the newly created log entry (e.g.
/xyz/openbmc_project/logging/entry/5) but createError() was discarding it. Change createError() to read and return that object path so callers can use it.

Add a new createBmcDump(bus, eventId) overload that accepts the log entry object path and passes it as CreateParameters::FilePath to Dump.Create.CreateDump(). The dump manager extracts FilePath and passes it verbatim to dreport as -p, which sets the $optional_path variable inherited by all dreport plugins. The dreport elog plugin then uses this path to collect the full error log entry details inside the dump archive. This is the same mechanism already used by elog_watch in phosphor-debug-collector.

The existing createBmcDump(bus) overload is preserved and delegates to the new overload with an empty string so all existing call sites remain unaffected. Callers that do not need the return value of createError() can discard it — C++ allows this without any changes at those sites.

Use the new API in host_state_manager when the host enters Quiesce state after exhausting boot retries as a demonstration.

Change-Id: Ia7e9b2ec044d170e244355fa7b216d81ea69ed10

### DIFF
--- a/host_state_manager.cpp
+++ b/host_state_manager.cpp
@@ -226,12 +226,14 @@ bool Host::isAutoReboot()
                 // Generate log since we will now be sitting in Quiesce
                 const std::string errorMsg =
                     "xyz.openbmc_project.State.Error.HostQuiesce";
-                utils::createError(this->bus, errorMsg,
-                                   sdbusplus::xyz::openbmc_project::Logging::
-                                       server::Entry::Level::Critical);
+                auto logEntryPath = utils::createError(
+                    this->bus, errorMsg,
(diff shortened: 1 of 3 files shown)


geissonator:
add to approvelist

